PR #3, opened 2018-07-04
Initial webhooks implementation

peterbroadhurst:
More to do to harden the webhook implementation (particularly providing a response path - whether a state store like MongoDB or post-back), but this currently provides a simple way to drive work into the Kafka bridge trivially over JSON (YAML needs a bit more work, but will be nice for Solidity deployments).


peterbroadhurst:
Quite a big structural change to the code in this branch, to split the KafkaBridge layer into two.
- `KafkaCommon` with the producer/consumer logic, common to Webhooks and the Bridge
- `KafkaBridge` the bridge itself
Also introduced a common set of mocks to use in both.

PR #19, opened 2018-07-26
Docs

peterbroadhurst:
First cut of the docs.

@Freyert - could you take a look and see what you think?


README.md
```diff
@@ -39,9 +41,56 @@ This code is distributed under the [Apache 2 license](LICENSE).
 
 > The code statically links to code distributed under the LGPL license.
 
+## Table of Contents
+
+- [github.com/kaleido-io/ethconnect](#githubcomkaleido-ioethconnect)
+  - [About kaleido-io/ethconnect](#about-kaleido-ioethconnect)
```
Freyert: I'm a huge fan of ToC so I really love this. I think removing the material that comes before the ToC would make the experience a little cleaner.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+### Ethereum Webhooks and the REST Receipt Store (MongoDB)
+
+Another key goal of having a robust Messaging layer under the covers is that the most trivial application can send messages into Ethereum reliably.
+
+- `POST` a [trivial YAML/JSON payload](#yaml-to-submit-a-transaction)
+  - to `/hook` to complete once the message is confirmed by Kafka (0.5s - tunable)
+  - to `/fasthook` to complete immediately when the message is sent to Kafka
+- Get an _immediate_ receipt
+   - No need to wait many seconds for a block to be cut
```
Freyert: As someone who has worked a teeny bit with ethereum the idea of an _immediate_ receipt is a little scary to me. This sounds like Kafka just says OK I got your stuff, which is very similar to sending a transaction into a node. The long, short of it is I don't _believe_ in this immediate receipt.

I imagine what this means is that Kafka confirms _it_ has received my transaction, but I will receive a webhook at a later date. Which is exciting! No retry polling (which you talk about later). It might be good to hit that point here instead of leaving it for later.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+   - No need to have complex retry handling
+
+This is ideal for Applications, Integration-as-a-Service (IaaS) platforms, and built-in SaaS integrations that support Webhooks for event submission.
+
+However, SaaS administrators and end-users still need to be able to track down what happened for a particular transaction, to diagnose problems, and correlate details of the Ethereum transaction with the business event.
```
Freyert: I think this troubleshooting section could be another subheader. It may warrant more detail.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+The management of this `nonce` pushes complexity back to the application tier - especially for horizontally scaled Enterprise applications sending many transactions using the same sender. By using an ordered Messaging stream to submit messages into the Ethereum network, many applications are able to delegate this complexity to kaleido-io/ethconnect.
+
+The kaleido-io/ethconnect bridge contains all the logic necessary to communicate with the node to determine the next nonce, and also to cope with multiple requests being in flight within the same block, for the same sender (including [with IBFT](https://github.com/ethereum/EIPs/issues/650#issuecomment-360085474)).
+
+If a sender needs to achieve exactly-once delivery of transactions (vs. at-least-once) it is still necessary to allocate the nonce within the application and pass it into kaleido-io/ethconnect in the payload.  This allows the sender to control allocation of nonces using its internal state store / locking.
```
Freyert: I would provide links to the definitions of `exactly-once` and `at-least-once`. Could use a markdown footnote style link. Love the precision here.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+Given many Enterprise scenarios involve writing hashed proofs of completion of an off-chain transaction to a shared ledger (vs. performing the actual transaction on the chain), at-least-once delivery is sufficient in a wide range of cases. Smart Contracts can simply re-store the proof.
```
Freyert: Why is this true?

Freyert: Is it true?

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+It provides a trivially simple REST API:
+- `GET` `/reply/a789940d-710b-489f-477f-dc9aaa0aef77` to look for an individual reply
+- `GET` `/replies` to list the replies
+  - Ordered by time _received_ (not the order submitted) - listing the newest first
+  - `limit` and `skip` query parameters can be used to paginate the results
+
+A capped collection can be used in MongoDB to limit the storage. For example to store only the last 1000 replies received.
+
+### Scale and message ordering - nonce management
```
Freyert: Nonce Management for scale and message ordering.

Nonce Management is the more interesting bit of the headline for me.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+The transaction pooling/execution logic within an Ethereum node is based upon the concept of a `nonce`, which must be incremented exactly once each time a transaction is submitted from the same Ethereum address. There can be no gaps in the nonce values, or messages build up in the `queued transaction` pool waiting for the gap to be filled. This allows for deterministic ordering of transactions sent by the same sender.
```
Freyert: How do we cope with gaps?

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+The kaleido-io/ethconnect bridge contains all the logic necessary to communicate with the node to determine the next nonce, and also to cope with multiple requests being in flight within the same block, for the same sender (including [with IBFT](https://github.com/ethereum/EIPs/issues/650#issuecomment-360085474)).
+
+If a sender needs to achieve exactly-once delivery of transactions (vs. at-least-once) it is still necessary to allocate the nonce within the application and pass it into kaleido-io/ethconnect in the payload.  This allows the sender to control allocation of nonces using its internal state store / locking.
+
+Given many Enterprise scenarios involve writing hashed proofs of completion of an off-chain transaction to a shared ledger (vs. performing the actual transaction on the chain), at-least-once delivery is sufficient in a wide range of cases. Smart Contracts can simply re-store the proof.
+
+## Why Kafka?
```
Freyert: I would put this in a FAQ in the Wiki.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+We selected Kafka as the first Messaging platform (and built a multi-tenant secured Kafka transport into the Kaleido platform), because Kafka has message ordering and scale characteristics that are ideally suited to the Ethereum transaction model:
+- Transactions can be sprayed across partitions, while retaining order of the transactions for a particular sender. Allowing independent and dynamic scaling of the application, kaleido-io/ethconnect bridge and Go-ethereum node components.
+- The modern replication based cloud-native and continuously available architecture is ideal for the Kaleido platform, and is likely to be a good fit for the modern Microservice architectures that are common in Blockchain projects.
+
+## Topic & Messages
+
+The topic usage is very simple:
+- One topic delivering messages into the bridge
```
Freyert: A diagram may be more useful here.

README.md
```diff
@@ -70,15 +119,279 @@ solidity: |-
...
+
+The MongoDB receipt store adds two additional fields, used to retrieve the entries efficient on the REST interface:
+```json
+{
+  "_id": "a789940d-710b-489f-477f-dc9aaa0aef77",
+  "receivedAt": 1532520942884
+}
+```
+
+## Running the Bridge
```
Freyert: I think it may be good to cut here, transform this into a short _contributing_ or `developing` section and then move all this development information into the Wiki.

I also think this is missing a lot of detail; which can come at a later date. Documenting all the configuration parameters would be _very_ awesome.

Freyert: Or even packaging this with a `docker-compose` set up. That would make it unbelievably easy to get a local development environment (or playground) and I think it could be done.

README.md
peterbroadhurst: Great question. I will explain how such a Smart Contract that just stores a proof can be made idempotent very easily, so that the same data can be submitted multiple times and the result of what is stored on the chain is the same.

README.md
peterbroadhurst: Let me add some more detail here... the point in Ethereum is that the system doesn't, it's the responsibility of the application code to fill in the gap. If you let ethconnect do it, we'll do our best to avoid a gap, and if one were to occur due to some timing restart scenario we'll fill it (but maintaining order wherever possible, even if there's resubmission due to at-least-once delivery). If you manage nonces yourself in the app for exactly-once delivery, then only your app knows what goes in the gap.

README.md
peterbroadhurst: Fair suggestion that we've cluttered the doc, but I've found wikis very painful in Git in the past. A bit of an information black-hole. I think this is pretty critical information for a new user, because one of the OSS AMQP providers (or even IBM MQ) might seem the obvious choice to many readers

README.md
peterbroadhurst: ... think the docker-compose setup, and Docker image pushing to Dockerhub, is an awesome suggestion... but probably a separate PR. If you're interested, that would be an awesome contribution 😄

peterbroadhurst:
Thank you @Freyert - great feedback. I'll work on an update.

peterbroadhurst:
@Freyert - mind taking another look?
Hopefully I've improved things in the docs, and in some cases I've added in some more code as it seemed clear writing the docs to address your comments there were some gaps.

README.md
```diff
@@ -172,13 +177,12 @@ Another key goal of having a robust Messaging layer under the covers is that the
 - `POST` a [trivial YAML/JSON payload](#yaml-to-submit-a-transaction)
   - to `/hook` to complete once the message is confirmed by Kafka (0.5s - tunable)
   - to `/fasthook` to complete immediately when the message is sent to Kafka
-- Get an _immediate_ receipt
-   - No need to wait many seconds for a block to be cut
-   - No need to have complex retry handling
+- Receive back an `id` for the request straight away
+   - Let the bridge do the retry polling to get the Ethereum receipt once a block is cut
```
Freyert: Receive a webhook once the receipt is received?

README.md
```diff
@@ -190,25 +194,54 @@ It provides a trivially simple REST API:
...
+sender to fill). This allows for deterministic ordering of transactions sent by the same sender.
 
 The management of this `nonce` pushes complexity back to the application tier - especially for horizontally scaled Enterprise applications sending many transactions using the same sender. By using an ordered Messaging stream to submit messages into the Ethereum network, many applications are able to delegate this complexity to kaleido-io/ethconnect.
 
 The kaleido-io/ethconnect bridge contains all the logic necessary to communicate with the node to determine the next nonce, and also to cope with multiple requests being in flight within the same block, for the same sender (including [with IBFT](https://github.com/ethereum/EIPs/issues/650#issuecomment-360085474)).
 
+When using the bridge, an application can submit simple YAML/JSON formatted transactions
```
Freyert: I'm assuming here that we are doing some very aggressive nonce checking? Checking the txpool before submitting new transactions and blocks?
What if instead of filling the nonce gap with new transactions my intention is to increase the gas expenditure on transactions stuck in the txpool so that miners will process them?

README.md
```diff
@@ -190,25 +194,54 @@ It provides a trivially simple REST API:
...
 
-Given many Enterprise scenarios involve writing hashed proofs of completion of an off-chain transaction to a shared ledger (vs. performing the actual transaction on the chain), at-least-once delivery is sufficient in a wide range of cases. Smart Contracts can simply re-store the proof.
+> There's a good summary of at-least-once vs. exactly-once semantics in the [Akka documentation](https://doc.akka.io/docs/akka/current/general/message-delivery-reliability.html?language=scala#discussion-what-does-at-most-once-mean-)
+
+Given many Enterprise scenarios involve writing hashed proofs of completion of an off-chain transaction to a shared ledger (vs. performing the actual transaction on the chain), at-least-once delivery is sufficient in a wide range of cases. Smart Contracts can be implemented in an idempotent way, to re-store or discard the proof if submitted twice.
+
+An `idempotent` interface is one that can be called multiple times with the same input,
```
Freyert: A nice throw in for the business users, or something to enable technicals to communicate with their business peers.

internal/kldutils/cmdutils.go
```diff
@@ -48,3 +51,17 @@ func DefInt(envVarName string, defValue int) int {
 	}
 	return int(parsedInt)
 }
+
+// MarshalToYAML marshals a JSON annotated structure into YAML, by first going to JSON
+func MarshalToYAML(conf interface{}) (yamlBytes []byte, err error) {
```
Freyert: I'm betting this becomes a hot piece of code at some point, but it gets the job done. Something to keep in the back of our heads to improve performance.

README.md
peterbroadhurst: Today we don't have way to do HTTP webhooks _back_ to the application. Instead we provide a store of transactions you query the status. That means HTTP polling.
So the trivially simple HTTP API, is just like the JSON/RPC API in that regards - you get an ID, and to know the success you have to poll.

If you want full async programming today, then that's available by coding to Kafka directly.

README.md
peterbroadhurst: At the moment the code is optimised for the Enterprise permissioned chain scenario, where there is no Ether economy for mining. We do not replace transactions that have been accepted into the pending transactions pool. When there's no mining economy these should not get stuck due to a miner deciding the gasPrice is too low (the gasPrice should be zero). However, if there is an unfunded account a transaction could get stuck there. There's no special code here to manage a transaction that blocks up a whole sender because the sender needs funding.

internal/kldutils/cmdutils.go
peterbroadhurst: It's only used at the moment for the `-Y` parameter, as a helper to dump the output.

There is similar code in the YAML->JSON part of the Webhooks bridge, which from much reading online seems to be the 'correct' way to do things with the inability to tag up a structure for _both_ YAML and JSON. The performance fix for users there is to submit JSON directly.

YAML's really just for humans using Postman/Curl to try things out.

PR #24, opened 2018-08-27
Default to letting the Node assign the nonce for node-signed txns

### LINKED ISSUE #21: Handle "replacement transaction underpriced" error in nonce management
The component handles nonce by calling getTransactionCount() for the submitting account to get the proper `nonce` value for the next transaction to submit. Sometimes this is not as reliable as can be (depending on the state of the pending transactions from the same account).

peterbroadhurst:
This PR changes the nonce prediction to be done on the node, rather than in the Go code, by default when submitting node-signed transactions.

This is to resolve #21 where we have seen that the `eth_getTransactionCount` JSON/RPC call is less reliable and then server-side in returning the most up to date nonce. This affects cases where ethconnect is not the only submitter of transactions to a node, and other clients (such as `web2` clients) are using the same accounts in the node wallet for signing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ Flags:
   -g, --consumer-group string    Client ID (or generated UUID)
   -h, --help                     help for kafka
   -m, --maxinflight int          Maximum messages to hold in-flight
-  -P, --predict-nonces           Predict the next nonce before sending txns (default=false for mode-signed txns)
+  -P, --predict-nonces           Predict the next nonce before sending txns (default=false for node-signed txns)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -359,6 +359,7 @@ Flags:
   -g, --consumer-group string    Client ID (or generated UUID)
   -h, --help                     help for kafka
   -m, --maxinflight int          Maximum messages to hold in-flight
+  -P, --predict-nonces           Predict the next nonce before sending txns (default=false for mode-signed txns)
```
jimthematrix: "node-signed"? ;-)

PR #26, opened 2018-11-08
Terminate when any goroutine fails in server mode

peterbroadhurst:
Currently when in `server` mode we wait for all goroutines to complete before we exit.
However, this means that if that just one of the server elements fail the process keeps running.
That is bad in a modern containerized environment as the container will keep limping along with half the functionality, rather that crash-recycling in the control of the container manager (kubernetes etc.).

### DIFF
--- a/cmd/ethconnect.go
+++ b/cmd/ethconnect.go
@@ -20,7 +20,6 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
-	"sync"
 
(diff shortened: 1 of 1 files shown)


jimthematrix:
both bridges' `Start()` method are blocking until they fail or crash, and would return to the main thread which causes the `true` signal to be sent through the channel. so either bridge failing will cause the process to terminate

PR #38, opened 2019-12-19
Add plugpoint for dynamic mapping from signing address to RPC endpoint

peterbroadhurst:


internal/kldtx/addressbook.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (ab *addressBook) resolveHost(endpoint string) (*url.URL, error) {
+	// We need to connect, but first we need to resolve the hostname to an IP
+	url, err := url.Parse(endpoint)
+	if err != nil {
+		log.Errorf("Invalid URL '%s': %s", endpoint, err)
+		return nil, fmt.Errorf("Invalid URL obtained for address")
+	}
+
+	if ab.conf.HostsFile != "" {
```
vdamle: For single region environments, we are not going to find the hosts file and so we use the endpoint URL as is, correct? (without translating to an IP address)

internal/kldtx/addressbook.go
```diff
@@ -0,0 +1,184 @@
...
+	url, err := ab.resolveHost(endpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// Connect and cache the RPC connection
+	rpc, err = kldeth.RPCConnect(&kldeth.RPCConnOpts{
+		URL: url.String(),
+	})
+	if err == nil {
```
vdamle: I think we should check `err != nil` and `return nil, err` here:

```
if err != nil {
  return nil, err
}
ab.hostToRPC[endpoint] = rpc
return rpc, nil
```
In the current form, we return the rpc connection even if there is an error

internal/kldutils/httprequester.go
```diff
@@ -0,0 +1,118 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// HTTPRequester perofrms common HTTP request logging/processing for utilities
```
vdamle:
```suggestion
// HTTPRequester performs common HTTP request logging/processing for utilities
```

PR #41, opened 2020-01-15
Support additional filter options on /replies

peterbroadhurst:
- `/replies?from=0x47F58d5b469799b17FC7c9D4B854af08F7e365Fc`
- `/replies?to=0x47F58d5b469799b17FC7c9D4B854af08F7e365Fc`
- `/replies?since=1579128548000` // millisecond epoch timestamp (exclusive greater than)
- `/replies?since=2020-01-15T22:30:00.00Z` // ISO8601 / RFC3339 format

### DIFF
--- a/internal/kldrest/receiptstore.go
+++ b/internal/kldrest/receiptstore.go
@@ -38,7 +38,7 @@ var uuidCharsVerifier, _ = regexp.Compile("^[0-9a-zA-Z-]+$")
 
 // ReceiptStorePersistence interface implemented by persistence layers
 type ReceiptStorePersistence interface {
-	GetReceipts(skip, limit int, ids []string, after, from, to string) (*[]map[string]interface{}, error)
+	GetReceipts(skip, limit int, ids []string, sinceEpochMS int64, from, to string) (*[]map[string]interface{}, error)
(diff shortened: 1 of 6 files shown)


internal/kldrest/receiptstore.go
```diff
@@ -194,7 +194,7 @@ func (r *receiptStore) getReplies(res http.ResponseWriter, req *http.Request, pa
 	}
 
 	// Call the persistence tier - which must return an empty array when no results (not an error)
-	results, err := r.persistence.GetReceipts(skip, limit, ids)
+	results, err := r.persistence.GetReceipts(skip, limit, ids, req.FormValue("since"), req.FormValue("from"), req.FormValue("to"))
```
vdamle: Not a big deal, but maybe better to move the input validation of `since` from `GetReceipts()` to here (since the other fields are validated here)

vdamle: I'm ok for us to do it in a subsequent change

peterbroadhurst: Thanks @vdamle - making this change. It's still a little clunky, because memory receipts don't have the filtering implemented (didn't seem worth it in the cost vs. benefit)

PR #45, opened 2020-01-30
HD Wallet support

peterbroadhurst:
> Note this delivers into the feature branch `accessToken` (which needs review)

### Primary change

- Allows `kld-from` and `from` field in JSON messages to have a new format:
  - `hd-instanceid-walletid-12345`
- Adds a new configuration section under processor (same level as addressbook)

```yaml
hdWallet:
  # Go template for URL to GET when accessing HD accounts
  urlTemplate: "https://someconstant-{{.InstanceID}}/api/v1/{{.WalletID}}/{{.Index}}"
  # ChainID for signing in ethconnect
  chainID: "12345"
  # Property names to look for in response
  propNames:
    address: "addr"
    privateKey: "key"
```

- Implements EIP155 signing in ethconnect based on remote accessed key
  - Note this requires secure connectivity to a REST HDWallet interface

### Other improvements

- Fixes `Makefile` to always build when source code changes
- Updates to Geth 1.9.10
  - Note `abi.NewType` signature has changed (again)
- Improves log formatting


internal/kldeth/send.go
```diff
@@ -155,13 +161,14 @@ type sendTxArgs struct {
 }
 
 // sendUnsignedTxn sends a transaction for internal signing by the node
-func (tx *Txn) sendUnsignedTxn(ctx context.Context, rpc RPCClient, txArgs *sendTxArgs) (string, error) {
+func (tx *Txn) sendUnsignedTxn(ctx context.Context, rpc RPCClient, txArgs *SendTXArgs) (string, error) {
```
vdamle: should this function now be renamed, since we `Sign()` the tx (if a `Signer` is present)?

PR #64, opened 2020-04-18
eth_call opts to query against historical block height

jimthematrix:
Add a `kld-opts` param to go along with `kld-call` to allow the request to target a historical block height as defined https://github.com/ethereum/wiki/wiki/JSON-RPC#the-default-block-parameter


internal/kldcontracts/rest2eth.go
```diff
@@ -190,6 +190,7 @@ type restCmd struct {
 	deployMsg *kldmessages.DeployContract
 	body      map[string]interface{}
 	msgParams []interface{}
+	option		string
```
vdamle, line 193: if we do go with `options` as the input, should this be a slice instead of a single string?

vdamle: rest of the internal calls seem to treat the input as variadic.. hence this looks out of place

internal/kldopenapi/abi2swagger.go
```diff
@@ -379,6 +391,7 @@ func (c *ABI2Swagger) addCommonParams(op *spec.Operation, isPOST bool, isConstru
 	privateForParam, _ := spec.NewRef("#/parameters/privateForParam")
 	privacyGroupIdParam, _ := spec.NewRef("#/parameters/privacyGroupIdParam")
 	registerParam, _ := spec.NewRef("#/parameters/registerParam")
+	optionParam, _ := spec.NewRef("#/parameters/optionParam")
```
vdamle, line 394: could be just me, I find `optionParam` vs `kld-opts` as query param kinda non-intuitive. I believe most other options try to maintain same name

jimthematrix:
This is going to use a `call option` parameter that's documented in the ethereum JSON-RPC spec for a number of methods besides `eth_call`:

| method | option |
|---------|-------|
| `eth_getbalance` | integer block number, or the string "latest", "earliest" or "pending" |
| `eth_getcode` | integer block number, or the string "latest", "earliest" or "pending" |
| `eth_getstorageat` | integer block number, or the string "latest", "earliest" or "pending" |
| `eth_gettransactioncount` | integer block number, or the string "latest", "earliest" or "pending" |
| `eth_call` | integer block number, or the string "latest", "earliest" or "pending" | 
| `eth_getblockbyhash` | If true it returns the full transaction objects, if false only the hashes of the transactions |
| `eth_getblockbynumber` | If true it returns the full transaction objects, if false only the hashes of the transactions |

Should be sufficient to make it a single string parameter, `kld-opts`, that can accommodate the above values. right now only using it for the `eth_call` downstream requests

internal/kldcontracts/rest2eth.go
jimthematrix: sorry making it variadic was mainly to not require that as a parameter, it'll only be a single parameter for all the different uses (see the new comment with the table of JSON RPC calls that uses such a "call option" parameter

internal/kldopenapi/abi2swagger.go
jimthematrix: now that I think about that, single it's always a single value parameter, `option` sounds better than `opts`. Now it should be consistent after the latest commit.

jimthematrix:
after further discussions with @vdamle and @peterbroadhurst, going with a dedicated parameter for the popular "target block number" parameter supported by a number of endpoints. Reserving the general purpose "option" or "opts" parameter for a future usage of more globally applicable parameter.

jimthematrix:
showing the test results with the update. The value stored onchain was changed from `10` to `20` at block 16226. 

Using the Gateway API to call the `get()` method of the smart contract, specifying `kld-blocknumber: 16224`:
image

Then specifying `kld-blocknumber: 16226`:
image

internal/kldeth/txn.go
```diff
@@ -121,13 +123,37 @@ func NewContractDeployTxn(msg *kldmessages.DeployContract, signer TXSigner) (tx
...
-func CallMethod(ctx context.Context, rpc RPCClient, signer TXSigner, from, addr string, value json.Number, methodABI *abi.Method, msgParams []interface{}) (map[string]interface{}, error) {
+func CallMethod(ctx context.Context, rpc RPCClient, signer TXSigner, from, addr string, value json.Number, methodABI *abi.Method, msgParams []interface{}, blocknumber string) (map[string]interface{}, error) {
 	log.Debugf("Calling method: %+v %+v", methodABI, msgParams)
 	tx, err := buildTX(signer, from, addr, "", value, "", "", methodABI, msgParams)
 	if err != nil {
 		return nil, err
 	}
-	retBytes, err := tx.Call(ctx, rpc)
+	callOption := "latest"
+	// only allowed values are "latest", "", a number string "12345" or a hex number "0xab23"
```
peterbroadhurst, line 133: looks like `pending`/`earliest` is supported below too

internal/kldeth/txn.go
```diff
@@ -121,13 +123,37 @@ func NewContractDeployTxn(msg *kldmessages.DeployContract, signer TXSigner) (tx
...
+	// "latest" and "" (no kld-blocknumber given) are equivalent
+	if blocknumber != "" && blocknumber != "latest" {
+		isHex, _ := regexp.MatchString(`^0x[0-9a-fA-F]+$`, blocknumber)
+		if isHex || blocknumber == "earliest" || blocknumber == "pending" {
+			callOption = blocknumber
+		} else {
+			n := new(big.Int)
+	    n, ok := n.SetString(blocknumber, 10)
+	    if !ok {
+	    	log.Warnf("Failed to convert blocknumber to big integer: %s", blocknumber)
```
peterbroadhurst, line 143: Surely we should return to the user at this point with a new defined error?

internal/kldeth/txn.go
```diff
@@ -121,13 +123,37 @@ func NewContractDeployTxn(msg *kldmessages.DeployContract, signer TXSigner) (tx
...
+	    	// when failed to convert, use "latest"
+	    } else {
+				bits := n.BitLen()
```
peterbroadhurst, line 146: Not sure I understand this bit. Can't we just use the `hexutil.Big` in the parameter? It's designed to do the serialization for us.

test/lotsoftypes.swagger.json
```diff
@@ -741,6 +750,12 @@
     }
   },
   "parameters": {
+    "blocknumberParam": {
+      "type": "string",
+      "description": "Used when kld-call is true, to specify the block number for the eth_call request to target. Empty or 'latest' to indicate the latest mined block. A number or a hex string to indicate a historical block height - 'x-kaleido-blocknumber' header can also be used",
```
peterbroadhurst, line 755: Think we need to do a bit of reconciliation of this text. Try and make them all short punchy lines, and defer to the docs for bigger explanation. Maybe compress the ` 'x-kaleido-blocknumber' header can also be used` on all the entries to `(header=x-kaleido-blocknumber)'

internal/kldeth/txn.go
jimthematrix: I agree that's better than handling it with default `latest`

internal/kldeth/txn.go
jimthematrix: since the end value we want is a hex string, once the input is detected as a hex it's used directly.

This section is for parsing integer strings into `big.Int`. Getting the bit length seems to be the best way to check for zero value

jimthematrix:
@peterbroadhurst thanks for the review, all comments should be addressed in the latest commit (and a response to your question on integer string handling)

PR #65, opened 2020-04-20
REST API viewer shows ":undefined" for type on empty inputs/outputs

peterbroadhurst:
Compiling a contract with methods that have no inputs or no outputs (such as a default constructor) the OpenAPI renders with a confusing `:undefined` output - because we do not include `"type": "object"` in the empty schema defintiion.

image


peterbroadhurst:
Added a fix for https://github.com/kaleido-io/ethconnect/issues/66 to this PR.

PR #67, opened 2020-04-20
Add tuning parameters for high concurrency submission of transactions

peterbroadhurst:
For the vast majority of use cases, single-threaded submission from Kafka partition -> node is desired - once parallelism has been obtained in the app tier.

e.g. the transactions are executed in the order they are submitted to each Kafka partition.

However, there are cases where the volume of data that needs to be submitted via a single ethereum node per block is so high, that the estimate + RLP encode + sign + send loop being single-threaded becomes a bottleneck.

This PR adds a set of new tuning options that solve for these high volume cases, noting that each comes with some trade-offs around ordering:

- `kafka.flush.frequency` - we used to wait 500ms by default for messages to arrive in the REST API before pushing to Kafka. This makes that tunable externally, and changes the default to be the much more friendly `0` for low volume use cases.
- `kafka.flush.messages` - number of messages to wait for before flushing when sending messages via the REST gateway - only applied when frequency > 0
- `kafka.flush.bytes` - number of bytes to wait for before flushing sending messages via the rEST gateway - only applied when frequency > 0
- `sendConcurrency` - how much parallelism to use when driving `eth_sendTransaction` calls (and the associated `eth_estimateGas`) calls to the node.
   > Note setting this higher than `maxInFlight` (which is the maximum number of messages we stream ahead of the acknowledged offset) does not have any effect.
- `alwaysManageNonces` - **existing** tuning parameter, but the nonce management code has been significantly revised.
  > If this is left to `false`, and `sendConcurrency > 1`, then _ordering is not assured_ because the `eth_sendTransaction` calls will be submitted in parallel to the Ethereum node, which has responsibility for assigning nonces. So setting `alwaysManageNonces=true` is recommended for `sendConcurrency > 1`.
- `attemptGapFill` - when dealing with parallelism, you always make a potential compromise on order. When `alwaysManageNonces=true` and `sendConcurrency > 1` ethconnect is careful to retain order _within a from address_ by assigning the nonces sequentially before submitting the transactions to the node. However, this causes a potential issue if one of the parallel transactions were to fail _before_ being submitted to the pending transaction pool. Detail below. This option allows you to auto-fill the gap that can be created by this situation, and hence allow later transactions to continue.

## The gap challenge

Circumstances that need to combine:

- A transaction is rejected by the node before entering the pending pool
    - A reason like `insufficient funds for gas * price + value`
- The nonce has been assigned because alwaysManageNonces=true`
- The are already transactions _for the same `from` address_ that have been allocated nonces ahead of this one
- We might want these later transactions be executed, because:
    - They don't suffer from the same problem as the earlier transaction
    - Their logic is not dependent on the outcome of the earlier transaction

In this case, we don't want to simply time out waiting for those transactions to be mined (they never will because it's a gap). Instead we want to fill the gap.

If `attemptGapFill=false`, then the behaviour is:
- The later transactions time out, unless there is another transaction ready and `sendConcurrency` is high enough to allow that to be submitted. In which case the _new transaction_ would be inserted in the earlier slot - so will be _executed out of order_.

If `attemptGapFill=true`, then the behaviour is:
- A nil transaction (eth transfer to/from the sender, with `value=0` and `gasPrice=0`) is submitted into the gap allowing all later transactions to be submitted.


peterbroadhurst:
This PR also adds new JSON fields to the error message that is stored in the asynchronous reply store, in the case that a gap-fill transaction was initiated by ethconnect:
```
gapFillTxHash
gapFillSucceeded
```

PR #71, opened 2020-06-17
Add Name to Events, CustomName to Subscription

vdamle:
* Adds `Name` to EventStream
* Adds `CustomName` to Subscription - there's already a system generated `Name`, which uses address and event signature
* Change `transactionIndex` in `logEntry` to be a hex uint instead of json `Number` to match https://github.com/ethereum/go-ethereum/blob/4bcc0a37ab70cb79b16893556cffdaad6974e7d8/core/types/log.go#L47. This was exposed by a behavior change in `go1.14` (Resulted in a corrupted slice) when parsing invalid `Number` in json (see commit for more details)
* Allow Solidity versions > `0.6.0` & < `0.6.9`

Partial fix for: https://github.com/kaleido-io/ethconnect/issues/68

### DIFF
--- a/internal/kldevents/subscription.go
+++ b/internal/kldevents/subscription.go
@@ -43,14 +43,14 @@ type ethFilter struct {
 // SubscriptionInfo is the persisted data for the subscription
 type SubscriptionInfo struct {
 	kldmessages.TimeSorted
-	ID         string                     `json:"id,omitempty"`
-	Path       string                     `json:"path"`
-	Name       string                     `json:"name"`
-	CustomName string                     `json:"customName,omitempty"`
-	Stream     string                     `json:"stream"`
-	Filter     persistedFilter            `json:"filter"`
-	Event      kldbind.MarshalledABIEvent `json:"event"`
-	FromBlock  string                     `json:"fromBlock,omitempty"`
+	ID        string                     `json:"id,omitempty"`
+	Path      string                     `json:"path"`
+	Summary   string                     `json:"-"`    // System generated name for the subscription
+	Name      string                     `json:"name"` // User provided name for the subscription, set to Summary if missing
+	Stream    string                     `json:"stream"`
+	Filter    persistedFilter            `json:"filter"`
+	Event     kldbind.MarshalledABIEvent `json:"event"`
+	FromBlock string                     `json:"fromBlock,omitempty"`
 }
 
 // subscription is the runtime that manages the subscription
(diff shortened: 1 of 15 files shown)


internal/kldevents/subscription.go
```diff
@@ -43,13 +43,14 @@ type ethFilter struct {
...
-	Path      string                     `json:"path"`
-	Name      string                     `json:"name"`
-	Stream    string                     `json:"stream"`
-	Filter    persistedFilter            `json:"filter"`
-	Event     kldbind.MarshalledABIEvent `json:"event"`
-	FromBlock string                     `json:"fromBlock,omitempty"`
+	ID         string                     `json:"id,omitempty"`
+	Path       string                     `json:"path"`
+	Name       string                     `json:"name"`
+	CustomName string                     `json:"customName,omitempty"`
```
peterbroadhurst, line 49: I think it's going to be confusing to have `customName` here.
I'm guessing the problem was we auto-generate the `name` field previously from combination of the event signature and the scope - which is useful information regardless of whether a user has a custom name.

With that in mind, please can we make the _new_ field `summary`, and have that always be auto-generated (cannot be user-submitted - works just like `name` does today). Then `name` defaults to the value of `summary` if it's empty/null, but if supplied the user can put anything they like?

internal/kldtx/txnprocessor_test.go
```diff
@@ -605,10 +605,13 @@ func TestOnSendTransactionMessageBadNonce(t *testing.T) {
 
 	txnProcessor := NewTxnProcessor(&TxnProcessorConf{}, &kldeth.RPCConf{}).(*txnProcessor)
 	testTxnContext := &testTxnContext{}
+	// Guessing this is a go version behavior change: using `abc` for nonce throws an error in parsing the
```
peterbroadhurst, line 608: This is a super helpful review comment. However, I'm not sure we want this in the code forever.
Maybe move it into the PR commentary?

internal/kldtx/txnprocessor_test.go
```diff
@@ -630,7 +630,7 @@ func TestOnSendTransactionMessageBadMsg(t *testing.T) {
 		"  \"headers\":{\"type\": \"SendTransaction\"}," +
 		"  \"from\":\"0x83dBC8e329b38cBA0Fc4ed99b1Ce9c2a390ABdC1\"," +
 		"  \"nonce\":\"123\"," +
-		"  \"value\":\"abc\"," +
+		"  \"value\":\"123.456\"," +
```
vdamle, line 633: Guessing this is a go version behavior change: using `abc` for nonce throws an error in parsing the json in the test file: "json: invalid number literal, trying to unmarshal "\"abc\"" into Number" instead of throwing an error in txn processing logic

internal/kldevents/subscription.go
vdamle: Added a new field `summary` (not returned in API response). `name` is now set to whatever is provided by the user or is set to `summary` if it is empty.

PR #73, opened 2020-06-29
Update EventStream (http PATCH)

vdamle:
* All fields in the event stream except the ID are allowed to be updated
* Event streams could be updated while batches of events are currently being processed. The existing logic around concurrent handling of batch processing of events and any pause/stopping of event streams ensures that batch processing exits as soon as a pause/stop event is detected. This PR adds an additional empty struct channel that can be closed to signal that the update of the stream is underway. The handler go routines select on the `updateInterrupt` ch so that they are woken up when the update is underway and can exit. The actual update to the event stream spec is performed after the go routines exit. Subsequently, a fresh round of event handler go routines is started.
* Fixes a bug that was incorrectly updating inflight count even if stream is suspended/stopped. Check that we processed the event before decrementing the in-flight counter
* Adds command to generate html for coverage

Addresses one of the items in: https://github.com/kaleido-io/ethconnect/issues/68

### DIFF
--- a/internal/kldcontracts/smartcontractgw.go
+++ b/internal/kldcontracts/smartcontractgw.go
@@ -676,6 +676,10 @@ func (g *smartContractGW) updateStream(res http.ResponseWriter, req *http.Reques
 		return
 	}
 	newSpec, err := g.sm.UpdateStream(req.Context(), streamID, &spec)
+	if err != nil {
+		g.gatewayErrReply(res, req, err, 500)
+		return
+	}
 
 	status := 200
 	log.Infof("<-- %s %s [%d]", req.Method, req.URL, status)
--- a/internal/kldcontracts/smartcontractgw.go
+++ b/internal/kldcontracts/smartcontractgw.go
@@ -115,6 +115,7 @@ func (g *smartContractGW) AddRoutes(router *httprouter.Router) {
 	router.GET("/gateways/:gateway_lookup", g.getRemoteRegistrySwaggerOrABI)
 	router.GET("/g/:gateway_lookup", g.getRemoteRegistrySwaggerOrABI)
 	router.POST(kldevents.StreamPathPrefix, g.withEventsAuth(g.createStream))
+	router.PATCH(kldevents.StreamPathPrefix, g.withEventsAuth(g.updateStream))
 	router.GET(kldevents.StreamPathPrefix, g.withEventsAuth(g.listStreamsOrSubs))
 	router.GET(kldevents.SubPathPrefix, g.withEventsAuth(g.listStreamsOrSubs))
 	router.GET(kldevents.StreamPathPrefix+"/:id", g.withEventsAuth(g.getStreamOrSub))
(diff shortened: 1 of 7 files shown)


internal/kldevents/eventstream.go
```diff
@@ -90,6 +90,9 @@ type eventStream struct {
 	batchCount        uint64
 	initialRetryDelay time.Duration
 	backoffFactor     float64
+	updateInProgress  bool
+	updateInterrupt   chan struct{}
```
peterbroadhurst, line 94: think it's best to be explicit on the type here

internal/kldevents/eventstream.go
```diff
@@ -272,22 +354,34 @@ func (a *eventStream) eventPoller() {
 				}
 			}
 		}
-		time.Sleep(a.pollingInterval)
+		// the event poller reacts to notification about a stream update, else it starts
+		// another round of polling after completion of the pollingInterval
+		select {
+		case <-a.updateInterrupt:
```
peterbroadhurst, line 360: Without an error catch here, won't this panic when the stream is closed?

peterbroadhurst: Or just make it a `bool` stream, and push a boolean into it?

internal/kldevents/eventstream.go
```diff
@@ -304,16 +398,30 @@ func (a *eventStream) batchDispatcher() {
 					return
 				}
 				currentBatch = append(currentBatch, event)
+			case <-a.updateInterrupt:
```
peterbroadhurst, line 401: Same note here as above on closing

internal/kldevents/eventstream.go
```diff
@@ -304,16 +398,30 @@ func (a *eventStream) batchDispatcher() {
...
+				// we were notified by the caller about an ongoing update, cancel the timeout ctx and return
+				log.Infof("%s: Notified of an ongoing stream update, will not dispatch batch", a.spec.ID)
+				a.updateWG.Done()
```
peterbroadhurst, line 404: Should this just be a `defer`? Does this fulfil the same purpose as `dispatcherDone`

internal/kldevents/eventstream.go
peterbroadhurst: I guess there might be multiple consumers on it, so `close` is an option that allows the same channel to be used in multiple places

internal/kldevents/eventstream.go
vdamle: Using a zero-sized `struct{}` to implement an alternative to `context.Context` is a recommendation I found in multiple blogs. ex:

https://www.ardanlabs.com/blog/2017/10/the-behavior-of-channels.html

`If you choose to use your own channel for cancellation, rather than the context package, your channel should be of type chan struct{}. It is the zero-space, idiomatic way to indicate a channel used only for signalling.`

peterbroadhurst: Makes sense - so it's defining an empty struct (rather than generic)

internal/kldevents/eventstream.go
peterbroadhurst: Discussed on the phone. We've confirmed in a `select` you don't need to explicitly add the second parameter, in order to get triggered with `nil`. If this did a panic, then the tests would be failing.

internal/kldevents/eventstream.go
vdamle: i'll add a comment to make that point, i agree the intent isn't clear as-is

internal/kldcontracts/smartcontractgw.go
```diff
@@ -654,6 +655,37 @@ func (g *smartContractGW) createStream(res http.ResponseWriter, req *http.Reques
...
+	if err != nil {
+		g.gatewayErrReply(res, req, err, 404)
+		return
+	}
+	var spec kldevents.StreamInfo
+	if err := json.NewDecoder(req.Body).Decode(&spec); err != nil {
+		g.gatewayErrReply(res, req, klderrors.Errorf(klderrors.RESTGatewayEventStreamInvalid, err), 400)
+		return
+	}
+	newSpec, err := g.sm.UpdateStream(req.Context(), streamID, &spec)
```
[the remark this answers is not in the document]
vdamle: added an error check and return `500` if so

internal/kldevents/eventstream.go
vdamle: ref: https://github.com/kaleido-io/ethconnect/pull/73/files#r447236954

internal/kldcontracts/smartcontractgw.go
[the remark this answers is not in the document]
vdamle: resolved in latest commit

internal/kldevents/eventstream.go
vdamle: moved this and other calls to a `defer`. Left invocation in `batchProcessor` as is since we need to unlock after calling `Done()` and that's not in all cases before returning

PR #74, opened 2020-07-07
Track and use highest nonce among in-flight transactions

vdamle:
* The current logic for assigning nonce runs through the in-flight list to
determine the maximum and assigns one higher
* When a batch of transactions are in-flight and a receipt for the txn
with highest nonce is received before all others which have lower
nonces, the txn is cleared from the in-flight list. The next
transaction from the same account is re-assigned the nonce that was used
by the cleared transaction, which results in a "nonce too low" error.
This PR introduces a `highestNonce` that is maintained for the in-flight
batch and is updated whenever a new transaction is added. This way, any
transactions which are cleared do not negatively impact the calculation
of the nonce when the next transaction is processed.

### DIFF
--- a/internal/kldtx/hdwallet_test.go
+++ b/internal/kldtx/hdwallet_test.go
@@ -71,6 +71,7 @@ func TestHDWalletSignOK(t *testing.T) {
 	s, err := hd.SignerFor(hdr)
 	assert.NoError(err)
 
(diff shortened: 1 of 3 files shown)


vdamle:
Good catch @peterbroadhurst . Agree with both points, the stale nonce can be really problematic. Will fix this.

PR #79, opened 2020-07-30
Update dependencies handling changes to geth 1.9.18

peterbroadhurst:
As a pre-req to supporting `ABICoderV2` we need to update our go-ethereum dependency.
There are some significant changes that affected this to the type interface.

Particularly as covered here

- Removal of `ID()` methods
- Requirement to use the constructor to generate signatures on a Method (affects order of our code significantly)
- Need to handle `stateMutability` field on the ABI

Also in Sarama, there is a new `RefreshController` method so I regenerated the `gomock`s


peterbroadhurst:
I also added a goreport badge - as https://github.com/kaleido-io/ethconnect/pull/33 has been languishing as I keep coming back to it, and thinking it needs updating. So the simple badge would move us forwards.

internal/kldeth/txn.go
```diff
@@ -372,9 +384,20 @@ func genMethodABI(jsonABI *kldmessages.ABIMethod) (method *abi.Method, err error
...
+		outputs[i] = arg
 	}
-	return
+
+	var abiMethod abi.Method
+	// Note we only support "normal" functions here - constructors do not follow this code path, and we don't have support
+	// of Fallback or Receive yet
+	switch jsonABI.Type {
+	case "function":
+	case "":
```
vdamle, line 395: Is this for `Fallback` function? If so, the comment preceding this is contradictory

peterbroadhurst: Actually, no. This is just to cover the (likely) case that when specifying the ABI in-line in YAML, the user does not include a `type` in there.

PR #80, opened 2020-07-31
ABIEncoderV2 support for struct (tuple) function input/output

peterbroadhurst:
Updates the core JSON <-> ABI logic of ethconnect to handle `ABIEncoderV2` encoded structs.

- Updates error messages to include the path to the field that is in error, including array indexes and nesting
- Tested with nested structures and arrays
- Validated fields are optional on structures
- Exploits all the same type mapping of the individual fields to/from JSON that top-level fields have
- Also includes some format and spelling fixes identified by go report.

Additional work likely to be required around OpenAPI (Swagger) generation, and potentially events.

### DIFF
--- a/internal/klderrors/errors.go
+++ b/internal/klderrors/errors.go
@@ -333,27 +333,29 @@ const (
 	// TransactionSendBadGasPrice a user-supplied gasPrice (eth to pay for each unit of gas spent) string in the JSON input cannot be processed
 	TransactionSendBadGasPrice = "Converting supplied 'gasPrice' to big integer"
 	// TransactionSendInputTypeBadNumber the input JSON value supplied for a method parameter cannot be converted to a number
-	TransactionSendInputTypeBadNumber = "Method '%s' param %d: Could not be converted to a number"
+	TransactionSendInputTypeBadNumber = "Method '%s' param %s: Could not be converted to a number"
 	// TransactionSendInputTypeBadJSONTypeForNumber the input JSON value supplied for a method parameter was not a number or a string, and needs to be converted to a number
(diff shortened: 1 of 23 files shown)


internal/klderrors/errors.go
```diff
@@ -367,7 +369,9 @@ const (
 	// TransactionSendMsgTypeUnknown we got a JSON message into the core processor (from Kafka, Webhooks etc.) that we don't understand
 	TransactionSendMsgTypeUnknown = "Unknown message type '%s'"
 	// TransactionSendInputTooManyParams more parameters provided than specified on ABI
-	TransactionSendInputTooManyParams = "Supplied %d paramters for ABI that supports %d"
+  TransactionSendInputTooManyParams = "Supplied %d paramters for ABI that supports %d"
+  // TransactionSendInputNotAssignable if we end up in a situation where the generated type cannot be assiend
```
vdamle, line 373:
```suggestion
  // TransactionSendInputNotAssignable if we end up in a situation where the generated type cannot be assigned
```

internal/kldeth/txn.go
```diff
@@ -22,7 +22,8 @@ import (
 	"reflect"
 	"regexp"
 	"strconv"
-	"strings"
+  "strings"
```
vdamle, line 25: just a nit, this formatting looks wacky.

vdamle: this file probably needs a `gofmt` run

peterbroadhurst: I've run it through gofmt 👍

PR #81, opened 2020-08-04
More work on ABIEncoderV2 to account for missing Marshal implementations

peterbroadhurst:
So I found a pretty big issue, that took a while to work around - and sadly quite a lot more code change.

The `abi.ABI` interface and its children in go-ethereum provided `Unmarshal`, but does not provide `Marshal`.
Before we had worked around that by providing my own custom `Marshal` interface, that simply to the external fields of the go-ethereum objects and serialized them.

Sadly when the structures were updated to add the `components` and `internalType` extensions, these fields in their original form get thrown away during the `Unmarshal`:

https://github.com/ethereum/go-ethereum/blob/f5382591874220287de253bfc08b10afd5244927/accounts/abi/argument.go#L36-L60

This meant I had to find a solution to allow the code to serialize and de-serialize the original JSON, while also no re-implementing all of the complex type implementation that is super useful within the go-ethereum `abi.Type` interface.

I settled on moving all the code to using our own simple `ABIMarshaling` structure that represents the raw JSON, then just in time conversion into `abi.Method` / `abi.Event` when we want the full type structures.
Even with this compromise the change was quite extensive.

I also found some gotchas when testing end-to-end with real nested input output structures, and addressed those

### DIFF
--- a/internal/kldbind/typeutils.go
+++ b/internal/kldbind/typeutils.go
@@ -127,5 +127,4 @@ func ABIElementMarshalingToABIMethod(m *ABIElementMarshaling) (method *ABIMethod
 			method = &nMethod
 		}
 	}
-	return
 }
--- a/internal/kldbind/typeutils.go
+++ b/internal/kldbind/typeutils.go
@@ -90,7 +90,12 @@ func ABIArgumentsMarshalingToABIArguments(marshalable []ABIArgumentMarshaling) (
 	arguments := make(ABIArguments, len(marshalable))
 	var err error
 	for i, mArg := range marshalable {
-		arguments[i].Type, err = abi.NewType(mArg.Type, mArg.InternalType, mArg.Components)
+		var components []abi.ArgumentMarshaling
+		if mArg.Components != nil {
+			b, _ := json.Marshal(&mArg.Components)
+			json.Unmarshal(b, &components)
+		}
+		arguments[i].Type, err = abi.NewType(mArg.Type, mArg.InternalType, components)
 		if err != nil {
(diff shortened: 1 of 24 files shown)


internal/kldbind/typeutils.go
```diff
@@ -74,108 +74,58 @@ func ABIEventSignature(event *ABIEvent) string {
...
+			method = &nMethod
 		}
-		abiArgs = append(abiArgs, ABIArgument{
-			Name:    arg.Name,
-			Type:    t,
-			Indexed: arg.Indexed,
-		})
 	}
-	return abiArgs, nil
+	return
```
vdamle, line 130:
```suggestion
```

internal/kldeth/txn.go
```diff
@@ -147,34 +152,26 @@ func CallMethod(ctx context.Context, rpc RPCClient, signer TXSigner, from, addr
...
+	var err error
+	if unpackErr != nil {
+		err = klderrors.Errorf(klderrors.UnpackOutputsFailed, unpackErr)
+	} else {
+		err = processOutputs(args, rawRetval, retval)
 	}
-	if err = processOutputs(args, rawRetval, retval); err != nil {
-		addErrorToRetval(retval, retBytes, rawRetval, err)
+	if err != nil {
+		log.Errorf("Unabled to process bytes '%s': %s", rawRetval, err)
```
vdamle, line 172:
```suggestion
		log.Errorf("Unable to process bytes '%s': %s", rawRetval, err)
```

PR #83, opened 2020-08-04
Fix how ABI serializes down into Kafka message

peterbroadhurst:
Now that we have hierarchical data structures being written into the intermediate message, we have to move the serialization structures across.

Due to the change made in #81 we lost a previous half-way house where we were serializing the type directly into the message alongside the value, and setting `methodName` (rather than the full `method` ABI). That approach does not scale to the deeply nested structures we need to now support.


peterbroadhurst:
going to remove the intermediate `kldmessages.ABIArgument` / `kldmessages.ABIMethod` fields as confusing and unnecessary

PR #84, opened 2020-08-04
Add noauth and schemes options to swagger gen

peterbroadhurst:
Allows tweaking of the swagger that is generated in the backend, to allow disabling of UI controls that Swagger UI would otherwise automatically add to its generated UX.

### DIFF
--- a/internal/kldcontracts/smartcontractgw.go
+++ b/internal/kldcontracts/smartcontractgw.go
@@ -859,7 +859,16 @@ func (g *smartContractGW) isSwaggerRequest(req *http.Request) (swaggerGen *kldop
 			conf.BasicAuth = strings.ToLower(vs[0]) == "false"
 		}
 		if vs := req.Form["schemes"]; len(vs) > 0 {
-			conf.ExternalSchemes = strings.Split(vs[0], ",")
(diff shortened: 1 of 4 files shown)


internal/kldcontracts/smartcontractgw.go
```diff
@@ -829,24 +834,35 @@ func (g *smartContractGW) resolveAddressOrName(id string) (deployMsg *kldmessage
...
+		abiRequest = strings.ToLower(vs[0]) != "false"
 	}
 	from = req.FormValue("from")
+	if swaggerRequest {
+		var conf = *g.baseSwaggerConf
+		if vs := req.Form["noauth"]; len(vs) > 0 {
+			conf.BasicAuth = strings.ToLower(vs[0]) == "false"
+		}
+		if vs := req.Form["schemes"]; len(vs) > 0 {
+			conf.ExternalSchemes = strings.Split(vs[0], ",")
```
vdamle, line 862: Should we filter out any schemes other than `http/https` here, with a warning?

peterbroadhurst: good point - injection potential

PR #87, opened 2020-08-12
FIx error parsing events with empty data

peterbroadhurst:
When an event only has indexed fields, and no data entries, the following error was observed: 

```
ERROR Failed to process event: sb-c4ad11c7-e1e9-4878-78b6-f23a874fb9fa:CampaignDeployed(string,uint256): Failed to parse RLP data from event: Expected nil in JSON/RPC response. Received: []
```

This has always been an issue, but there has been a change in the latest version from previous versions.

Previously we used to emit the event with the following embedded error:
```
{
  "address": "0x19E75d0d337e17835dc5246f007A1fB17f0bAC89",
  "blockNumber": "475266",
  "transactionIndex": "0x0",
  "transactionHash": "0x23307094299f08a1041de9f1e7ecb67197a5a3c11ce5be775a8147de266b7524",
  "data": {
    "var1": "0x51b201b016025d42c9a0718b75aacc12b1e9c7f16e4bd2c6618aa944ca399156",
    "error": "Expected nil in JSON/RPC response. Received: []",
    "raw": [],
    "rlp": "",
    "supply": "1000"
  },
  "subId": "sb-432cf526-7be6-48e2-5e2f-14d15daee5ee",
  "signature": "ExampleEvent(string,uint256)",
  "logIndex": "0"
}
```

### DIFF
--- a/internal/kldeth/txn.go
+++ b/internal/kldeth/txn.go
@@ -16,6 +16,7 @@ package kldeth
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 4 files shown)


peterbroadhurst:
The change in how we process the error is in https://github.com/kaleido-io/ethconnect/commit/ffeff016cdccc5af17abfcbb6ed93af6752b2341

See https://github.com/kaleido-io/ethconnect/commit/ffeff016cdccc5af17abfcbb6ed93af6752b2341#r41394132

peterbroadhurst:
Looking to re-instate previous serialization of errors

peterbroadhurst:
ok - the scope of the behavior change was:
- `GET` (`eth_call`) results starting to return an error return code, rather than an embedded `"error"` structure
- Events where the RLP could not be decoded - where the most common was this example of an empty event.

PR #88, opened 2020-08-17
Add POST /reset route for subscriptions, with fromBlock

peterbroadhurst:
To allow a subscription to be reset back to an earlier block, and replay all those events from the chain

Addresses an enhancement in: https://github.com/kaleido-io/ethconnect/issues/68

### DIFF
--- a/internal/kldevents/submanager.go
+++ b/internal/kldevents/submanager.go
@@ -56,6 +56,7 @@ type SubscriptionManager interface {
 	AddSubscription(ctx context.Context, addr *kldbind.Address, event *kldbind.ABIElementMarshaling, streamID, initialBlock, name string) (*SubscriptionInfo, error)
 	Subscriptions(ctx context.Context) []*SubscriptionInfo
 	SubscriptionByID(ctx context.Context, id string) (*SubscriptionInfo, error)
+	ResetSubscription(ctx context.Context, id, initialBlock string) error
 	DeleteSubscription(ctx context.Context, id string) error
 	Close()
 }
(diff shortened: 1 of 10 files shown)


internal/kldevents/submanager.go
```diff
@@ -125,6 +126,20 @@ func (s *subscriptionMGR) Subscriptions(ctx context.Context) []*SubscriptionInfo
 	return l
 }
 
+func (s *subscriptionMGR) setInitialBlock(i *SubscriptionInfo, initialBlock string) error {
+	// Check initial block number to subscribe from
+	if initialBlock == "" || initialBlock == FromBlockLatest {
```
vdamle, line 131: Can we add a log message here?

internal/kldevents/submanager.go
```diff
@@ -125,6 +126,20 @@ func (s *subscriptionMGR) Subscriptions(ctx context.Context) []*SubscriptionInfo
...
+		i.FromBlock = FromBlockLatest
+	} else {
+		var bi big.Int
+		if _, ok := bi.SetString(initialBlock, 0); !ok {
```
vdamle, line 135: Why are we using base `0` to test validity and base `10` to set `FromBlock`?

internal/kldevents/submanager.go
vdamle: never mind, i see we are logging `initialBlock` in `requestReset()`

internal/kldevents/submanager.go
peterbroadhurst: The first part *sets* the value from the input string, with a friendly way of base `0` - meaning `0x1234` would work as hex, and `1234` would work as decimal.

However, the `FromBlock` is then consistently stored as a base-10 string. So the serialized format is _always_ base 10, regardless of what the user submitted

PR #90, opened 2020-09-09
ABI: Handle unnamed input parameters in functions

vdamle:
* Solidity supports un-named parameters in functions. The ABI parsing
logic is now enhanced to assign a default name that is prefixed `input`
when processing such functions with unnamed parameters. If there are
multiple unnamed parameters, a suffix with the parameter's index is
added to the name. API requests shall have to use the above nomenclature
when sending requests.
* Request parsing logic is also updated to handle unnamed parameters in
the ABI and look for default names in the query parameters, as described
above.


internal/kldcontracts/rest2eth_test.go
```diff
@@ -963,6 +963,40 @@ func TestSendTransactionDefaultConstructorABI(t *testing.T) {
...
+	abiLoader := &mockABILoader{
+		deployMsg: &kldmessages.DeployContract{
+			ABI: kldbind.ABIMarshaling{
+				{
+					Name: "unnamedparamsmethod", Type: "function", Inputs: []kldbind.ABIArgumentMarshaling{
+						{Name: "", Type: "uint256", InternalType: "uint256"},
+						{Name: "", Type: "uint256", InternalType: "uint256"},
+					},
+				},
+			}, // completely empty ABI is ok
```
vdamle, line 985:
```suggestion
			},
```

internal/kldcontracts/rest2eth.go
```diff
@@ -419,12 +420,21 @@ func (r *rest2eth) resolveParams(res http.ResponseWriter, req *http.Request, par
 	c.msgParams = make([]interface{}, len(c.abiMethod.Inputs))
 	queryParams := req.Form
 	for i, abiParam := range c.abiMethod.Inputs {
-		if bv, exists := c.body[abiParam.Name]; exists {
+		argName := abiParam.Name
+		// If the ABI input has one or more un-named parameters, look for default names that are passed in.
+		// Unnamed Input params should be named: input, input1, input2...
```
peterbroadhurst, line 425: Going to dig around for precedent here in the code - I think we have a naming convention on start from 0 or 1 etc. that it's worth being consistent with...

peterbroadhurst: Aha, this is what we already have in `genOutput` 👍

peterbroadhurst: (bet it was me who set that precedent... feels a bit odd in retrospect, but ah well 🤷‍♂️ )

peterbroadhurst (approved):
📥 📥 📥 
0️⃣ 1️⃣ 2️⃣

PR #92, opened 2020-09-15
Add tests to demonstrate valid/invalid cases for bytesxx input

vdamle:


peterbroadhurst (approved):
👍

peterbroadhurst:
One thought is that we could do with translating https://github.com/kaleido-io/ethconnect/blob/8444b2b5de87cf7c9ef85249d48e34fa41835504/internal/kldeth/txn.go#L539 into a human readable table that describes the map interface.

PR #95, opened 2020-10-06
WebSocket support for event streams

peterbroadhurst:
## Why vanilla WebSockets not Socket.io?

I investigated all of the socket.io server implementations, and actually got to 90% on coding up the implementation with the most robust and active one: https://github.com/googollee/go-socket.io, but I failed when I got to the end to end testing.

The conclusion was that there isn't an up to date socket.io 2.0 compliant server engine, with good test coverage, and an active OSS community.
Also the general swing of opinion seems to be away from socket.io to vanilla WebSockets.

## The WebSocket API

Route: `/ws` - same listener as HTTP/HTTPS API

## Topics

In order to allow multiple event streams to target different sets of WebSocket applications, I've introduced a lightweight `topic` construct. This is optionally specified on the configuration of the `eventstream` (the default topic is `""`), and  then when the application listens or acks it can specify the topic.

### Listen

```js
{
  "type": "listen",
  "topic": "mytopic"
}
```

### Ack

The model currently _only_ has acknowledgement of batches, and you must acknowledge a batch before the next batch will be delivered. This makes it equivalent to the REST API, and means we don't flood the application.

```js
{
  "type": "ack",
  "topic": "mytopic"
}
```

### Error (nack)

The application can choose to nack a payload, if it cannot process it. That will result in the standard exponential backoff retry and blocking behavior kicking in on the server. This means the app logic can stay super simple when using websockets, and all the retry stays in the server side.

```js
{
  "type": "error",
  "topic": "mytopic",
  "message": "My backend is down right now... give me a minute will ya?"
}
```

## Architecture

The `kldws` package elevates the low-level WebSocket API to provide the above model to the other components. For now that's just Event Streams, but it can be extended trivially to include receipt notifications for async TX submission.

The `WebSocketChannels` interface contains:

```go
GetChannels(topic string) (sender chan<- interface{}, receiver <-chan error)
```

This lets the `eventstreams` assert the existence of a `topic` that it has in its configuration, by asking for a bidirectional set of channels. Then it can block sending messages into the `sender` channel, and any active connections that have `listen`ed to that `topic` will compete to pull messages from it. If none are available, the `sender` will block indefinitely.

> Note that the client can also assert the existence of a `topic` - so we don't care which comes first.

The `eventstreams` interface does an interruptible indefinite send on this topic, meaning:
- A batch is not considered `blocked` just because no-one is connected to the `topic`
- If the stream is updated, then it will wake up as expected


peterbroadhurst:
I've done end to end testing of this now.

jimthematrix (approved):
LGTM

PR #96, opened 2020-10-07
Couple of event stream fixes

peterbroadhurst:
- Allow editing of the `Topic` for a WebSocket event stream
- Prevent changing the type of an Event Stream
- Bail before editing properties when the user has an error in their config for update
- Set a more sensible default number of seconds for the request timeout - 120seconds rather than 30,000seconds


jimthematrix (approved):
👍

PR #99, opened 2020-10-30
Do not wipe existing eventstream name when updating it

vdamle:
Check that the update request has a valid name before overwriting
existing name


peterbroadhurst (approved):
👍

PR #103, opened 2020-12-11
Reset to checkpoint after evenstream update

peterbroadhurst:
If you perform an update of an event stream while a batch is in-flight, then the batch will not be replayed after the update.
That's because the update logic exits all the handlers immediately.
If you stop and restart ethconnect, it will replay the events. Also, if more events arrive these are delivered - if you ack them, then that moves the checkpoint to a HWM beyond the original batch, so those earlier events are missed.

For Webhooks this is a small timing window, but for WebSockets it's large - because a batch stays in-flight while there are no websockets clients connected. So if you do an update while your app is down, and there are messages waiting for it, then those messages get missed.

fyi @gabriel-indik

### DIFF
--- a/internal/kldevents/eventstream.go
+++ b/internal/kldevents/eventstream.go
@@ -405,7 +405,7 @@ func (a *eventStream) eventPoller() {
 		case <-a.updateInterrupt:
 			// we were notified by the caller about an ongoing update, no need to continue
 			log.Infof("%s: Notified of an ongoing stream update, exiting event poller", a.spec.ID)
-			a.markAllSubscripitonsStale(ctx)
+			a.markAllSubscriptionsStale(ctx)
 			return
 		case <-time.After(a.pollingInterval): //fall through and continue to the next iteration
 		}
--- a/internal/kldevents/eventstream.go
+++ b/internal/kldevents/eventstream.go
@@ -411,7 +411,7 @@ func (a *eventStream) eventPoller() {
 		}
 	}
 
-	a.markAllSubscripitonsStale(ctx)
+	a.markAllSubscriptionsStale(ctx)
 
 }
 
--- a/internal/kldevents/eventstream.go
+++ b/internal/kldevents/eventstream.go
@@ -325,7 +325,7 @@ func (a *eventStream) isBlocked() bool {
 	return v
 }
 
-func (a *eventStream) markAllSubscripitonsStale(ctx context.Context) {
+func (a *eventStream) markAllSubscriptionsStale(ctx context.Context) {
 	// Mark all subscriptions stale, so they will re-start from the checkpoint if/when we re-run the poller
 	subs := a.sm.subscriptionsForStream(a.spec.ID)
 	for _, sub := range subs {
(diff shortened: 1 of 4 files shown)


internal/kldevents/subscription.go
```diff
@@ -257,3 +249,18 @@ func (s *subscription) requestReset() {
 func (s *subscription) blockHWM() big.Int {
 	return s.lp.getBlockHWM()
 }
+
+func (s *subscription) markFilterStale(ctx context.Context, newFilterStale bool) {
+	log.Debugf("%s: Marking filter stale=%t", s.logName, newFilterStale)
```
vdamle, line 254:
```suggestion
	log.Debugf("%s: Marking filter stale=%t, current sub filter stale=%t", s.logName, newFilterStale, s.filterStale)
```

internal/kldevents/eventstream.go
```diff
@@ -328,10 +328,17 @@ func (a *eventStream) isBlocked() bool {
...
-	defer a.updateWG.Done()
-
 	ctx := kldauth.NewSystemAuthContext()
 
+	defer func() {
+		a.updateWG.Done()
+		// Mark all subscriptions stale, so they will re-start from the checkpoint if/when we re-run the poller
+		subs := a.sm.subscriptionsForStream(a.spec.ID)
+		for _, sub := range subs {
+			sub.markFilterStale(ctx, true)
```
vdamle, line 338: There's one thing to be aware of, now that we are doing a few extra things in this deferred function (which potentially involve RPC calls to uninstall filters). Basically what I was thinking is that because deferred functions are stacked, we signal the event poller as done first up, then signal the wait group and then go through marking all the subscriptions stale. It seems like depending on the number of filters that exist in the stream, uninstalling filters could take "some" time. Not sure if it will cause any issues when the stream is restarted soon after (I'm assuming that the filters are re-created).

peterbroadhurst: aha, that's a great point. Will refactor

peterbroadhurst: I've made the update 👍

internal/kldevents/eventstream.go
```diff
@@ -397,10 +405,14 @@ func (a *eventStream) eventPoller() {
 		case <-a.updateInterrupt:
 			// we were notified by the caller about an ongoing update, no need to continue
 			log.Infof("%s: Notified of an ongoing stream update, exiting event poller", a.spec.ID)
+			a.markAllSubscripitonsStale(ctx)
```
vdamle, line 408:
```suggestion
			a.markAllSubscriptionsStale(ctx)
```

internal/kldevents/eventstream.go
```diff
@@ -397,10 +405,14 @@ func (a *eventStream) eventPoller() {
...
 			return
 		case <-time.After(a.pollingInterval): //fall through and continue to the next iteration
 		}
 	}
+
+	a.markAllSubscripitonsStale(ctx)
```
vdamle, line 414:
```suggestion
	a.markAllSubscriptionsStale(ctx)
```

PR #104, opened 2021-01-06
Adjust highest nonce when there is no gap to fill

vdamle:
When a transaction cannot be sent due to an error (ex: gas estimation
fails), we check and see if a gap exists (inflight txns with higher
nonce exist), in which case a gap fill transaction is sent. However, in
the case where no gap exists, the `highest nonce` for the `from` address
must be reset to one less than the inflight transaction's nonce so that
the same nonce is re-used for the next incoming transaction.

### DIFF
--- a/internal/kldtx/txnprocessor.go
+++ b/internal/kldtx/txnprocessor.go
@@ -349,24 +349,25 @@ func (p *txnProcessor) cancelInFlight(inflight *inflightTxn, gapPotential bool) 
 				}
 			}
 		}
+		if higherNonceInflight < 0 {
+			// We did not find a higher nonce in-flight, so there's no gap to fill. However, we need to make sure
+			// that this nonce is re-used by the next incoming transaction for this address, if this is not the first transaction for the address.
+			// If this is the first txn, we would have cleared the key from the map - so there's nothing to update.
+			if gapPotential && p.inflightTxns[inflight.from] != nil {
+				log.Infof("Did not find a Transaction with higher nonce in-flight, setting highest nonce for %s to %d", inflight.from, inflight.nonce-1)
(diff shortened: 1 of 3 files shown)


peterbroadhurst (commented):
This makes perfect sense to me @vdamle 
I've not marked approval as it's in draft, and I don't know if you're planning to add a test case for this scenario before finalizing.

internal/kldtx/txnprocessor.go
```diff
@@ -349,6 +349,15 @@ func (p *txnProcessor) cancelInFlight(inflight *inflightTxn, gapPotential bool)
 				}
 			}
 		}
+		if higherNonceInflight < 0 {
+			// We did not find a higher nonce in-flight, so there's no gap to fill. However, we need to make sure
+			// that this nonce is re-used by the next incoming transaction for this address, if this is not the first transaction for the address.
+			// If this is the first txn, we would have cleared the key from the map - so there's nothing to update.
+			if gapPotential && p.inflightTxns[inflight.from] != nil {
+				log.Infof("Did not find a Transaction with higher nonce in-flight, setting highest nonce for %s to %d", inflight.from, inflight.nonce-1)
+				p.inflightTxns[inflight.from].highestNonce = inflight.nonce - 1
```
peterbroadhurst, line 358: Looks good. Reviewed the code above to confirm in the case we were the last transaction in flight for this nonce, we would have cleared the entry from `inflightTxns` already. So the additional nil check on the map catches that.

peterbroadhurst (approved):
So I believe the affected situation for this issue is:
- `alwaysManageNonce` or HD Wallet external signing
- enough load to drive multiple transactions from the same `from` address in per block
- a mixture of successful and failing transactions
- the nature of the failure is pre-submission to the node, such as gas estimation

PR #109, opened 2021-02-16
Treat every cancel as a potential gap, until confirmed sent

peterbroadhurst:
The gap fill logic currently assumes that if we didn't get as far as `eth_send` then there is no potential for `cancelnFlight` to perform gap management in the in-flight pool.

This incorrectly bypassed the fix in https://github.com/kaleido-io/ethconnect/pull/104, if the failure is during transaction preparation before calling `eth_send`. In that case, the action was not a gap-fill, but simply to update the `highestNonce` in the in-flight transaction pool.

I also found the code a little hard to read while investigating the issue, so did a little refactoring to aid readability, and updated the logging in case a further investigation is required.

### DIFF
--- a/internal/kldtx/txnprocessor.go
+++ b/internal/kldtx/txnprocessor.go
@@ -344,11 +344,9 @@ func (p *txnProcessor) cancelInFlight(inflight *inflightTxn, submitted bool) {
 			delete(p.inflightTxns, inflight.from)
 		} else if !submitted {
 			// Check the transactions that are left, to see if any nonce is higher
-			if !inflight.nodeAssignNonce {
-				for _, alreadyInflight := range inflightForAddr.txnsInFlight {
-					if alreadyInflight.nonce > highestNonce {
-						highestNonce = alreadyInflight.nonce
-					}
+			for _, alreadyInflight := range inflightForAddr.txnsInFlight {
+				if alreadyInflight.nonce > highestNonce {
+					highestNonce = alreadyInflight.nonce
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


internal/kldtx/txnprocessor.go
```diff
@@ -324,48 +324,47 @@ func (p *txnProcessor) addInflightWrapper(txnContext TxnContext, msg *kldmessage
...
 	}
 	p.inflightTxnsLock.Unlock()
 
-	log.Infof("In-flight %d complete. nonce=%d addr=%s before=%d after=%d", inflight.id, inflight.nonce, inflight.from, before, after)
+	log.Infof("In-flight %d complete. nonce=%d addr=%s nan=%t sub=%t before=%d after=%d highest=%d", inflight.id, inflight.nonce, inflight.from, inflight.nodeAssignNonce, submitted, before, after, highestNonce)
 
 	// If we've got a gap potential, we need to submit a gap-fill TX
-	if higherNonceInflight > 0 {
-		log.Warnf("Potential nonce gap. Nonce %d failed to send. Nonce %d in-flight", inflight.nonce, higherNonceInflight)
+	if !submitted && highestNonce > inflight.nonce && !inflight.nodeAssignNonce {
```
vdamle, line 366: ok, this final check and submit simplifies the earlier branches and makes the code more readable! 👍🏼

vdamle (approved):

peterbroadhurst:
I generated a simple recreate, that caused transaction timeouts so I can verify the fix:

```js
const axios = require('axios');

async function run() {

  const apikey = 'zzdxszj70e';
  const apisecret = 'xxxx';
  const nodeURL = 'xxx-yyy.kaleido.io';
  const contractAddr = '0x6aa156f36acb5fd361784ddb1eaa8a2dab1f6f92';
  const from = '0x1cb9041b976964c93455026e0643b94942d24ee6';
  const url = `https://${apikey}:${apisecret}@${nodeURL}/instances/${contractAddr}/transfer?kld-from=${from}&kld-sync=false`

  for (let amount = 0; amount < 10; amount++) {
    const promises = [];
    for (let i = 0; i < 25; i++) {
      promises.push((async () => {
        let payload = i % 5 === 0 ? {
          amount,
          recipient: "0xc000"
        } : {
          amount,
          recipient: "0xc03b1cdbfff3c43786bfcf3518703ba4908dc87f"
        };
        await axios.post(url, payload);
        console.log(`${i}: Sent ${JSON.stringify(payload)}`)
      })());
    }
    await Promise.all(promises);
  }

}

run().catch(err => { console.error(err.stack, err.response && err.response.data); process.exit(1); });
```

PR #110, opened 2021-04-22
Fix context for getTransactionReceipt in webhooksDirect mode (no Kafka)

nguyer:
Previously, we would pass the current request context to `GetTXReceipt()`. When not using Kafka, this does not work because the context is passed down from the original HTTP request. By the point we get here in the code, we have already responded to the HTTP request, so the context gets canceled. Passing a context that has already been canceled to the RPC package causes an immediate failure (https://pkg.go.dev/github.com/ethereum/go-ethereum@v1.9.23/rpc?utm_source=gopls#Client.CallContext). This change uses a background context so it is not related to the original HTTP request anymore.

### DIFF
--- a/internal/kldtx/txnprocessor.go
+++ b/internal/kldtx/txnprocessor.go
@@ -15,6 +15,7 @@
 package kldtx
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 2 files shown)


internal/kldtx/txnprocessor.go
```diff
@@ -406,7 +407,7 @@ func (p *txnProcessor) waitForCompletion(inflight *inflightTxn, initialWaitDelay
 	var elapsed time.Duration
 	for !isMined && !timedOut {
 
-		if isMined, err = inflight.tx.GetTXReceipt(inflight.txnContext.Context(), p.rpc); err != nil {
+		if isMined, err = inflight.tx.GetTXReceipt(context.Background(), p.rpc); err != nil {
```
peterbroadhurst, line 410: I think rather than changing the context for this one RPC call, it's potentially safer to update the creation of the context.
I believe you saw this because you're using the `webhooksDirect` implementation (rather than Kafka), and so probably it's better to swap to `context.Background()` in this `msgContext`:
https://github.com/kaleido-io/ethconnect/blob/3d441aeceb8be565fd38e044f197e82e6dd18237/internal/kldrest/webhooksdirect.go#L145

peterbroadhurst (changes requested):

internal/kldtx/txnprocessor.go
nguyer: That makes sense! I will update the PR.

peterbroadhurst (approved):
🌇 🌆 🌃
👩‍💼 👩‍💼 👩‍💼

PR #112, opened 2021-04-29
Fix websocket EventStream with no topic

nguyer:
When a websocket EventStream is created, if no topic is specified, the `spec` field is a `nil` pointer. This causes `ethconnect` to crash when trying to emit an event. Worse, because the event is persisted to the DB, it tries to send the event again when it starts back up again, causing it to enter an indefinite crash loop.

 This fix changes the existing behavior so that if no topic is specified, the topic is implicitly set to `""`.

### DIFF
--- a/internal/kldcontracts/smartcontractgw.go
+++ b/internal/kldcontracts/smartcontractgw.go
@@ -650,7 +650,6 @@ func (g *smartContractGW) createStream(res http.ResponseWriter, req *http.Reques
 		g.gatewayErrReply(res, req, klderrors.Errorf(klderrors.RESTGatewayEventStreamInvalid, err), 400)
 		return
 	}
-	log.Warn(spec.WebSocket)
 
 	newSpec, err := g.sm.AddStream(req.Context(), &spec)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/kldcontracts/smartcontractgw.go
```diff
@@ -650,6 +650,7 @@ func (g *smartContractGW) createStream(res http.ResponseWriter, req *http.Reques
 		g.gatewayErrReply(res, req, klderrors.Errorf(klderrors.RESTGatewayEventStreamInvalid, err), 400)
 		return
 	}
+	log.Warn(spec.WebSocket)
```
peterbroadhurst, line 653: Is this deliberately left, or lint from previous debugging?

peterbroadhurst (approved):
Fix looks great, and thanks for spotting and fixing those typos.

One comment to address before merging, but approval marked.

internal/kldcontracts/smartcontractgw.go
nguyer: Good catch. That was left from debugging, sorry. I missed that when I reviewed the diff. I will remove.

PR #113, opened 2021-04-29
Add receipts to mongo receipt store with retry

vdamle:
* To address transient failures writing receipts to mongo, add a retry with backoff.
* I've picked an initial delay of 1 second, a backoff factor of 1.1 and a max retry timeout of 120 seconds.
* also added requestID as an arg to AddReceipt() so that logs can include it as a prefix

### DIFF
--- a/internal/kldrest/mongoreceipts.go
+++ b/internal/kldrest/mongoreceipts.go
@@ -74,12 +74,29 @@ func (m *mongoReceipts) connect() (err error) {
 }
 
 // AddReceipt processes an individual reply message, and contains all errors
-func (m *mongoReceipts) AddReceipt(receipt *map[string]interface{}) error {
-	if err := m.collection.Insert(*receipt); err != nil {
-		log.Errorf("Insert failed for object '%.v': %s", *receipt, err)
-		return err
+// To account for any transitory failures writing to mongoDB, it retries adding receipt with a backoff
+func (m *mongoReceipts) AddReceipt(requestID string, receipt *map[string]interface{}) (err error) {
+	retryTimeout := 10 * time.Second
+	backoffFactor := 1.1
+	startTime := time.Now()
+	delay := 500 * time.Millisecond
+	attempt := 0
+	complete := false
+
+	for !complete {
+		if attempt > 0 {
+			log.Infof("%s: Waiting %.2fs before re-attempt:%d mongo write", requestID, delay.Seconds(), attempt)
+			time.Sleep(delay)
+			delay = time.Duration(float64(delay) * backoffFactor)
+		}
+		attempt++
+		err = m.collection.Insert(*receipt)
+		complete = err == nil || time.Since(startTime) > retryTimeout
+		if !complete {
+			log.Errorf("%s: addReceipt attempt: %d failed, err: %s, retryTimeout: %d", requestID, attempt, err, retryTimeout)
+		}
 	}
-	return nil
+	return err
 }
 
 // GetReceipts Returns recent receipts with skip & limit
--- a/internal/kldrest/mongoreceipts.go
+++ b/internal/kldrest/mongoreceipts.go
@@ -25,6 +25,8 @@ import (
 
 const (
 	mongoConnectTimeout = 10 * 1000
+	retryTimeout        = 120 * 1000
+	backoffFactor       = 1.1
 )
 
 type mongoReceipts struct {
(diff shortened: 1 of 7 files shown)


internal/kldrest/mongoreceipts.go
```diff
@@ -74,12 +74,29 @@ func (m *mongoReceipts) connect() (err error) {
...
+		if attempt > 0 {
+			log.Infof("%s: Waiting %.2fs before re-attempt:%d mongo write", requestID, delay.Seconds(), attempt)
+			time.Sleep(delay)
+			delay = time.Duration(float64(delay) * backoffFactor)
+		}
+		attempt++
+		err = m.collection.Insert(*receipt)
+		complete = err == nil || time.Since(startTime) > retryTimeout
+		if !complete {
+			log.Errorf("%s: addReceipt attempt: %d failed, err: %s, retryTimeout: %d", requestID, attempt, err, retryTimeout)
```
peterbroadhurst, line 96: Feel like probably we should panic here @vdamle - what do you think?

internal/kldrest/receiptstore.go
```diff
@@ -115,10 +115,10 @@ func (r *receiptStore) processReply(msgBytes []byte) {
 
 	// Insert the receipt into MongoDB - captures errors
 	if requestID != "" && r.persistence != nil {
-		if err := r.persistence.AddReceipt(&parsedMsg); err != nil {
-			log.Errorf("Failed to insert '%s' into receipt store: %+v", parsedMsg, err)
+		if err := r.persistence.AddReceipt(requestID, &parsedMsg); err != nil {
+			log.Errorf("%s: Failed to insert '%s' into receipt store: %+v", requestID, parsedMsg, err)
```
peterbroadhurst, line 119: Think we should probably panic here @vdamle - what do you think?

internal/kldrest/mongoreceipts.go
```diff
@@ -74,12 +74,29 @@ func (m *mongoReceipts) connect() (err error) {
 }
 
 // AddReceipt processes an individual reply message, and contains all errors
-func (m *mongoReceipts) AddReceipt(receipt *map[string]interface{}) error {
-	if err := m.collection.Insert(*receipt); err != nil {
-		log.Errorf("Insert failed for object '%.v': %s", *receipt, err)
-		return err
+// To account for any transitory failures writing to mongoDB, it retries adding receipt with a backoff
+func (m *mongoReceipts) AddReceipt(requestID string, receipt *map[string]interface{}) (err error) {
+	retryTimeout := 10 * time.Second
```
peterbroadhurst, line 79: I would suggest 120 secs as the timeout here.
Also, maybe move these to a `const(` section at the top of the file?

peterbroadhurst (commented):
Couple of questions @vdamle

internal/kldrest/receiptstore.go
vdamle: I missed this, sorry. Will add that

internal/kldrest/mongoreceipts.go
```diff
@@ -74,12 +76,30 @@ func (m *mongoReceipts) connect() (err error) {
...
 
 // AddReceipt processes an individual reply message, and contains all errors
-func (m *mongoReceipts) AddReceipt(receipt *map[string]interface{}) error {
-	if err := m.collection.Insert(*receipt); err != nil {
-		log.Errorf("Insert failed for object '%.v': %s", *receipt, err)
-		return err
+// To account for any transitory failures writing to mongoDB, it retries adding receipt with a backoff
+func (m *mongoReceipts) AddReceipt(requestID string, receipt *map[string]interface{}) (err error) {
+	startTime := time.Now()
+	delay := 500 * time.Millisecond
```
peterbroadhurst, line 82: the initial retry delay could move up to const too I think

internal/kldrest/receiptstore.go
```diff
@@ -115,10 +115,10 @@ func (r *receiptStore) processReply(msgBytes []byte) {
 
 	// Insert the receipt into MongoDB - captures errors
 	if requestID != "" && r.persistence != nil {
-		if err := r.persistence.AddReceipt(&parsedMsg); err != nil {
-			log.Errorf("Failed to insert '%s' into receipt store: %+v", parsedMsg, err)
+		if err := r.persistence.AddReceipt(requestID, &parsedMsg); err != nil {
+			log.Panicf("%s: Failed to insert '%s' into receipt store: %+v", requestID, parsedMsg, err)
```
peterbroadhurst, line 119: Have you confirmed this log statement causes the panic itself?

peterbroadhurst: Confirmed from logrus docs 👍

peterbroadhurst (approved):
Approved and merging

PR #114, opened 2021-05-05
WebSocket EventStream Broadcast

nguyer:
This PR adds a new behavior for WebSocket EventStreams. The new behavior can be enabled by creating an EventStream with the following request body:

```json
{
    "type": "websocket",
    "websocket": {
         "topic": "foobar",
         "distributionMode": "broadcast"
     }
}
```

The existing behavior is called `workloadDistribution`. It is a pseudo random distribution of Events over all open WebSocket connections listening on that particular topic. Each Event must be ack'd by the WebSocket client before the next event will be sent. If the `websocket` object or the `distributionMode` field is not specified, it will default to using the existing behavior of `workloadDistribution`.

The new behavior is called `broadcast`. It will send the same Event object to every open WebSocket connections listening on that particular topic, at the same time. If there are no open connections listening on that topic at the time of the broadcast, the Event will be dropped. Broadcast events should not be ack'd.

This PR also adds the ability to receive transaction replies (receipts) on a WebSocket broadcast. To begin receiving transaction replies a WebSocket client must send a message to the server indicating that they are interested in receiving replies.

```json
{
    "type": "listenReplies"
}
```

Transaction replies are sent to any and all open WebSocket connections that are listening (by sending the above message). If no WebSockets are currently listening for replies, the reply is not sent on a WebSocket. This will not block transaction operation. If the user is interested in a reply, but didn't have a WebSocket listening at the time that it was broadcast, it is still persisted to MongoDB where it can be retrieved later.

### DIFF
--- a/internal/kldevents/eventstream.go
+++ b/internal/kldevents/eventstream.go
@@ -115,6 +115,13 @@ type eventStreamAction interface {
 	attemptBatch(batchNumber, attempt uint64, events []*eventData) error
 }
 
+func validateWebSocket(w *webSocketActionInfo) error {
+	if w.DistributionMode != "" && w.DistributionMode != DistributionModeBroadcast && w.DistributionMode != DistributionModeWLD {
+		return klderrors.Errorf(klderrors.EventStreamsInvalidDistributionMode, w.DistributionMode)
+	}
+	return nil
+}
+
 // newEventStream constructor verifies the action is correct, kicks
 // off the event batch processor, and blockHWM will be
 // initialied to that supplied (zero on initial, or the
@@ -174,11 +181,9 @@ func newEventStream(sm subscriptionManager, spec *StreamInfo, wsChannels kldws.W
 	case "websocket":
 
 		if spec.WebSocket != nil {
-			distributionMode := spec.WebSocket.DistributionMode
-			if distributionMode != "" && distributionMode != DistributionModeBroadcast && distributionMode != DistributionModeWLD {
-				return nil, klderrors.Errorf(klderrors.EventStreamsInvalidDistributionMode, spec.WebSocket.DistributionMode)
+			if err := validateWebSocket(spec.WebSocket); err != nil {
+				return nil, err
 			}
-			spec.WebSocket.DistributionMode = distributionMode
 		}
 
 		if a.action, err = newWebSocketAction(a, spec.WebSocket); err != nil {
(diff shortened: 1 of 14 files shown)


internal/kldevents/eventstream.go
```diff
@@ -245,7 +255,13 @@ func (a *eventStream) update(newSpec *StreamInfo) (spec *StreamInfo, err error)
 	}
 	if a.spec.Type == "websocket" && newSpec.WebSocket != nil {
 		a.spec.WebSocket.Topic = newSpec.WebSocket.Topic
+		distributionMode := strings.ToLower(newSpec.WebSocket.DistributionMode)
```
peterbroadhurst, line 258: Wonder if it's worth pulling this validation into a function, to avoid having the `distributionMode != "broadcast" && distributionMode != "workloaddistribution"` twice (and potentially forgetting to update one of them in the future)

internal/kldevents/websockets.go
```diff
@@ -33,35 +35,47 @@ func newWebSocketAction(es *eventStream, spec *webSocketActionInfo) (*webSocketA
...
-	var err error
-	select {
-	case err = <-receiver:
-		break
-	case <-w.es.updateInterrupt:
-		return klderrors.Errorf(klderrors.EventStreamsWebSocketInterruptedReceive)
-	case <-closing:
-		return klderrors.Errorf(klderrors.EventStreamsWebSocketInterruptedReceive)
+	// If we ever add more distribution modes, we may want to change this logic from a simple if statement
+	if w.spec.DistributionMode != "broadcast" {
```
peterbroadhurst, line 68: An enum might be good... like:

```
type DistributionMode string
const (
  DistributionModeBroadcast DistributionMode = "broadcast"
  DistributionModeWLM DistributionMode = "workloadDistribution"
)
````

File: internal/kldws/wsconn.go [hunk 3]
```diff
@@ -132,9 +142,11 @@ func (c *webSocketConnection) listen() {
 		log.Debugf("WS/%s: Received: %+v", c.id, msg)
 
 		t := c.server.getTopic(msg.Topic)
-		switch msg.Type {
+		switch strings.ToLower(msg.Type) {
```
peterbroadhurst, line 145: Nice 👍

internal/kldws/wsconn.go
```diff
@@ -132,9 +142,11 @@ func (c *webSocketConnection) listen() {
...
 		case "listen":
 			c.listenTopic(t)
+		case "listenreplies":
```
peterbroadhurst, line 148: This word feels a little awkward... not sure if there's a shorter word, can't immediately think of one though 🤷

peterbroadhurst (commented):
Couple of thoughts for consideration @nguyer - but all minor.
Will take a little bit more of a look later

peterbroadhurst (approved):
✉️ 📢

PR #115, opened 2021-05-08
Swallow attempts to insert a duplicate receipt

peterbroadhurst:
Since #113 we will detect errors inserting receipts after processing messages, and retry for a period of time before panicing.

However, in the design of ethconnect there is the possibility of duplicate receipts in an edge case.
Specifically, when we successfully process a message and send a kafka response for that message (which is processed by the receipt store), but then crash before we move our Kafka offset forwards beyond that message.
The message will be replayed from Kafka after the restart, and hence a duplicate receipt can be inserted.

The action we can take at this point is to log and discard the duplicate receipt.

### DIFF
--- a/internal/kldrest/receiptstore_test.go
+++ b/internal/kldrest/receiptstore_test.go
@@ -32,27 +32,34 @@ import (
 	"github.com/kaleido-io/ethconnect/internal/kldutils"
 )
 
-type mockReceiptErrs struct{ err error }
+type mockReceiptErrs struct {
+	getReceiptsErr error
+	getReceiptVal  *map[string]interface{}
+	getReceiptErr  error
+	addReceiptErr  error
+}
 
 func (m *mockReceiptErrs) GetReceipts(skip, limit int, ids []string, sinceEpochMS int64, from, to string) (*[]map[string]interface{}, error) {
-	return nil, m.err
+	return nil, m.getReceiptsErr
 }
 
 func (m *mockReceiptErrs) GetReceipt(requestID string) (*map[string]interface{}, error) {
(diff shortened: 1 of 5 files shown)


internal/kldrest/receiptstore_test.go
```diff
@@ -187,6 +199,29 @@ func TestReplyProcessorWithPeristenceErrorSwallows(t *testing.T) {
...
+	replyMsg := &kldmessages.TransactionReceipt{}
+	replyMsg.Headers.MsgType = kldmessages.MsgTypeTransactionSuccess
+	replyMsg.Headers.ID = kldutils.UUIDv4()
+	replyMsg.Headers.ReqID = kldutils.UUIDv4()
+	replyMsg.Headers.ReqOffset = "topic:1:2"
+	txHash := common.HexToHash("0x02587104e9879911bea3d5bf6ccd7e1a6cb9a03145b8a1141804cebd6aa67c5c")
+	replyMsg.TransactionHash = &txHash
+	replyMsgBytes, _ := json.Marshal(&replyMsg)
+
+	r.processReply(replyMsgBytes)
```
nguyer, line 222: It would be good to validate the correct behavior (the mock?) was called in this particular case. It would give me more confidence that the new behavior flows as expected.

nguyer (approved):
One minor suggestion on the new test. Other than that, looks good to me.

internal/kldrest/receiptstore_test.go
peterbroadhurst: 👍  - added a verification that the error had been thrown during the test.

PR #117, opened 2021-05-27
Updates to docker build

peterbroadhurst:


nguyer (approved):
Looks good and it builds on my machine. But we all know how much of an assurance that is...

PR #119, opened 2021-05-27
Move to dynamic linking for go-ethereum dependency

peterbroadhurst:
To avoid any confusion with the licensing position of this repo, this PR makes the following changes:
- Remove any static import of `go-ethereum` (LGPL licensed) code from code this repo
- Introduce a build check to avoid any future re-introduction of direct static import
- Perform dynamic loading of a `.so` built from https://github.com/kaleido-io/ethbinding (which is used for headers at build time, as required for any dynamic linking)
- Local builds require a `ethbinding` directory at the same level as the `ethconnect` directory
  - Facilitated by a `replace` in the `go.mod`
  - Allows concurrent development of the two codebases
- Docker builds use 
  - Removes the `replace` from `go.mod` and pulls do directly via `go get` to build it
- Includes some small README updates (more to come there to refresh the readme)

fyi @jimthematrix - taking the liberty of moving this to a full PR state, as I know you're unavailable today and the Hyperledger TSC committee were keen to have this moving fowards.


nguyer:
I've validated this PR builds a Docker image and I tested it with the latest code in firefly master. Using the CLI I can stand up a new stack with this version of ethconnect and everything looks good to me, with the one notable exception that the Firefly messages aren't showing up in the UI. I believe that problem is outside the scope or impact of this PR though, so should be safe to merge from that standpoint.

internal/eth/ethbinding.go
```diff
@@ -0,0 +1,67 @@
...
+		// If we are in a subdirectory of internal, then assume we're running unit tests
+		if strings.HasSuffix(cwd, "cmd") {
+			// Use the project root
+			bindingLocation = fmt.Sprintf("%s/../ethbinding.so", cwd)
+		} else if strings.HasSuffix(path.Dir(cwd), "internal") {
+			// Use the project root
+			bindingLocation = fmt.Sprintf("%s/../../ethbinding.so", cwd)
+		} else {
+			// Use the local directory
+			bindingLocation = fmt.Sprintf("%s/ethbinding.so", cwd)
```
vdamle, line 50: To clarify, for any Kaleido environments that will be upgraded, the container will find `ethbinding.so` in the local directory. Is that correct?

peterbroadhurst: Yes, that's right. This is handled in the Kaleido embedding of this component. The OSS Dockerfile has been updated in this PR to make that true.

PR #127, opened 2021-07-01
Added support for LevelDB based receipts store

jimthematrix:
This enhancement uses a local LevelDB for receipts store, in addition to the ability to use an external mongodb.

## Configuration
The configuration for the leveldb based receipt store:

```
rest:
  rest-gateway:
    ...
    leveldb:
      path: "/data/ethconnect/receiptsdb"
      queryLimit: 100
```
## Interface Design
The interface `ReceiptStorePersistence` has been enhanced to include a `start` parameter to help with efficient paging, in addition to `skip`.

```
type ReceiptStorePersistence interface {
	GetReceipts(skip, limit int, ids []string, sinceEpochMS int64, from, to, start string) (*[]map[string]interface{}, error)
	GetReceipt(requestID string) (*map[string]interface{}, error)
	AddReceipt(requestID string, receipt *map[string]interface{}) error
}
```
## Detailed Implementation Design
Because leveldb lacks indexing support, we have to build our own indexes in order to support the query parameters.

Each entry is saved under a unique key generated with [ULID](https://github.com/oklog/ulid/v2). A `z` prefix is added to the generated key so that the lookup keys are saved at the top of the range that will be traversed by the iterator going from `Last()` backwards (`Prev()`).

Optionally 3 indexing entries are also generated each supporting one of the query parameters:
- `since`: to support the timestamp based queries, the `receivedAt` values of each record is used to add an additional entry `receivedAt:<timestamp>:<lookupKey>`
- `from`: to support the sender based queries, the `from` values of each record is used to add an additional entry `from:<sender address>:<lookupKey>`
- `to`: to support the target contract based queries, the `to` values of each record is used to add an additional entry `to:<contract address>:<lookupKey>`

Every entry returned by the `GetReceipts()` method contains the extra property `_sequenceKey`, with the value being the lookup key. This allows the client to use the value as the `start` parameter of a subsequent query


internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,117 @@
...
+// AddReceipt processes an individual reply message, and contains all errors
+// To account for any transitory failures writing to mongoDB, it retries adding receipt with a backoff
+func (l *levelDBReceipts) AddReceipt(requestID string, receipt *map[string]interface{}) (err error) {
+	b, _ := json.MarshalIndent(&receipt, "", "  ")
+	return l.store.Put(requestID, b)
+}
+
+// GetReceipts Returns recent receipts with skip & limit
+func (l *levelDBReceipts) GetReceipts(skip, limit int, ids []string, sinceEpochMS int64, from, to string) (*[]map[string]interface{}, error) {
+	itr := l.store.NewIterator()
```
peterbroadhurst, line 50: @jimthematrix - sorry if I've missed it, but what ensures that this is in reverse addition order?
I'm not sure that the natural order of the `requestID` strings would do that automatically.

jimthematrix: good question, I was thinking about that as well. 

Given that the iteration order is by default based on key lexicon order, we need to set a different order using a comparator. I'm thinking to use `receivedAt` as the first level check, but since its precision is millisecond, we need a tie breaker of the `timeElapsed` which is nanosecond. This way we don't have to implement a synthetic index

jimthematrix: guess the Comparer only operates on keys, not values. so the above won't work. may have to think about using a composite key assembled from the two fields mentioned above

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if index >= skip {
+			val := itr.Value()
+			if len(val) > 0 {
+				// the index keys don't have values. if we have hit an entry with a value
+				// then we have reached the end of the index keys
+				break
+			}
+			// extract the request key from the index key
+			r, _ := regexp.Compile("z[0-9.]+:(.+)")
```
peterbroadhurst, line 84: This is really performance sensitive code I think - potentially running millions of times for a single query, where we have a request that misses all records.
We should probably use a `split` if we can safely.
If we can't, we should definitely at least pre-compile the regexp.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			val, err := l.store.Get(lookupKey)
+			if err != nil {
+				fmt.Printf("Failed to locate record by key %s\n", lookupKey)
+				continue
+			}
+			receipt := make(map[string]interface{})
+			err = json.Unmarshal(val, &receipt)
+			if err != nil {
+				fmt.Printf("Failed to decode stored receipt for request %s. This record is skipped in the returned list", itr.Key())
```
peterbroadhurst, line 101: Could we move these all over to `logrus` please.
There's no linter to protect against `fmt.Printf` at the moment, so it relies on a manual review to avoid these creeping in.

peterbroadhurst: (not commenting on all the items individually - but there are a number in the PR)

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			// extract the request key from the index key
+			r, _ := regexp.Compile("z[0-9.]+:(.+)")
+			compositeKey := itr.Key()
+			matches := r.FindStringSubmatch(compositeKey)
+			var lookupKey string
+			if len(matches) == 2 {
+				lookupKey = matches[1]
+			} else {
+				fmt.Printf("Failed to parse composite key %s\n", compositeKey)
```
peterbroadhurst, line 91: What should be the action here?
- return an error
- continue

peterbroadhurst: ... noting this logic is I think redundant with other comments

internal/rest/leveldbreceipts.go
peterbroadhurst: ... came back to this as I was making another comment. I think we are overcomplicating things, and again it will have a performance implication. Please see notes on line 57 above

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,157 @@
...
+	timeSeconds := (*receipt)["receivedAt"].(int64) + int64(timeElapsed)
+	timeStringSeconds := fmt.Sprintf("%d", timeSeconds)
+	timeStringNanoSeconds := fmt.Sprintf("%.9f", timeElapsed)
+	timeStringNanoSeconds = timeStringNanoSeconds[strings.Index(timeStringNanoSeconds, "."):]
+	timestamp := timeStringSeconds + timeStringNanoSeconds
+
+	// add "z" prefix so these entries come after the actual entries
+	// because for iteration we start from last backwards
+	compositeKey := fmt.Sprintf("z%s:%s", timestamp, requestID)
+	l.store.Put(compositeKey, nil)
```
peterbroadhurst, line 57: I believe if you swap the logic such that we put the _real_ record under the `compositeKey`, and store the composite key under the `requestID` then you can really simplify the `GetReceipts` logic below.

It would mean that a get-by-ID would need a second Get, which of course is the payoff.

So I'll leave it with you whether you make some of the suggested tweaks to the performance of the `GetReceipts` loop below (particularly avoiding the Regexp compile), or just do the indirected query lookup on the `GetReceipt` (thus eliminating the need for the Regexp or split).

peterbroadhurst (changes requested):
Hey @jimthematrix - sorry, got myself very confused in the end with this.

There's a specific request here to remove the `fmt.Printf` calls.

Also one more thing to consider, is whether it's simpler (and more performant) to just switch to a natively time-sortable ID in the generation of the ID itself. One example library with a lot of usage:
https://github.com/oklog/ulid

internal/rest/restgateway.go
```diff
@@ -275,6 +282,14 @@ func (g *RESTGateway) Start() (err error) {
 		if err = mongoStore.connect(); err != nil {
 			return
 		}
+	} else if g.conf.LevelDB.Path != "" {
```
peterbroadhurst, line 285: One more @jimthematrix, sorry.
This configuration could be confusing for users, because the `openapi` config also contains a level db.
Maybe you could provide an example of how the config would look when you've specified the:
- LevelDB for event streams state
- LevelDB for OpenAPI config
- LevelDB for receitps

... just to make sure it's clear reading the YAML.

peterbroadhurst:
> Hey @jimthematrix - sorry, got myself very confused in the end with this.
> 
> There's a specific request here to remove the `fmt.Printf` calls.
> 
> Also one more thing to consider, is whether it's simpler (and more performant) to just switch to a natively time-sortable ID in the generation of the ID itself. One example library with a lot of usage:
> https://github.com/oklog/ulid

One more library I found: https://github.com/segmentio/ksuid

internal/rest/leveldbreceipts.go
jimthematrix: Think the GetReceipts() always processing multiple rounds is more performance sensitive, so went with the swap to have the generated key mapped to the real record, and having the requestID mapped to the generated key

jimthematrix:
The latest commit adds better handling of the query parameters to avoid getting a query to iterate through large amount of entries:
- entries returned by `GetReceipts()` have the additional property `_sequenceKey`, which can be used in paging with the `start` parameter
- creates additional indexes for `from`, `to` and `receivedAt` properties in order to support query parameters `from`, `to` and `since`
- if `since` is present, the iterator will be applied a range with the matched `receivedAt` entry for the epoch value being the range limit
- if `ids` are present, use it to look up the individual entries matching the list, then apply the bounds if `start` and `since`. No iterator is used
- if `from` or `to` are present, use it to look up the matching entries with the corresponding indexes
- `limit` is defaulted to `50`

internal/kvstore/kvstore.go
```diff
@@ -67,9 +71,14 @@ func (k *levelDBKeyValueStore) Delete(key string) error {
 	return err
 }
 
-func (k *levelDBKeyValueStore) NewIterator() KVIterator {
+func (k *levelDBKeyValueStore) NewIterator(slice ...interface{}) KVIterator {
```
peterbroadhurst, line 74: Might be clearer to have a separate `NewIteratorRange` function, if that's a small change before we merge.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	defaultLimit int
+}
+
+func newLevelDBReceipts(conf *LevelDBReceiptStoreConf) (*levelDBReceipts, error) {
+	store, err := kvstore.NewLDBKeyValueStore(conf.Path)
+	if err != nil {
+		return nil, errors.Errorf(errors.ReceiptStoreLevelDBConnect, err)
+	}
+	t := time.Unix(1000000, 0)
+	entropy := ulid.Monotonic(rand.New(rand.NewSource(t.UnixNano())), 0)
```
peterbroadhurst, line 45: Was reading this in the Godoc:
```
Please note that rand.Rand from the math package is not safe for concurrent use. Instantiate one per long living go-routine or use a sync.Pool if you want to avoid the potential contention of a locked rand.Source as its been frequently observed in the package level functions.
```

peterbroadhurst: This conversation was useful: https://github.com/cnabio/cnab-go/pull/207/files#r424489120
Sounds like the most straight forward safe implementation is a simple mutex around the generation (or probably more easily, the `AddReceipt function

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+		idEntropy:    entropy,
+		defaultLimit: 50,
+	}, nil
+}
+
+// AddReceipt processes an individual reply message, and contains all errors
+// To account for any transitory failures writing to mongoDB, it retries adding receipt with a backoff
+func (l *levelDBReceipts) AddReceipt(requestID string, receipt *map[string]interface{}) (err error) {
+	// insert an entry with a composite key to track the insertion order
+	newId := ulid.MustNew(ulid.Timestamp(time.Now()), l.idEntropy)
```
peterbroadhurst, line 59: See note above about a mutex around this function

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	// add "z" prefix so these entries come after the lookup entries
+	// because for iteration we start from last backwards
+	lookupKey := fmt.Sprintf("z%s", newId)
+
+	b, _ := json.MarshalIndent(receipt, "", "  ")
+	l.store.Put(lookupKey, b)
+
+	// build the index for "from"
+	fromKey := fmt.Sprintf("from:%s:%s", (*receipt)["from"], lookupKey)
```
peterbroadhurst, line 68: Sorry for the amount of work needed to create the indexed filtering in LevelDB.
I did do a bit of research around how problematic having a huge number of keys is, and people were talking in the low billions. So I don't _think_ having 5 keys per receipt is a problem, and for backwards compatibility I don't think we can avoid it. However, I wonder if for new users (particularly FireFly driven usage) we should have a config option to disable rich searching.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	if len(ids) > 0 {
+		// use the list of IDs to search for the result entries
+		lookupKeysByIDs = l.getLookupKeysByIDs(ids, start, endKey)
+	}
+	var lookupKeysByFromAndTo []string
+	if from != "" || to != "" {
+		lookupKeysByFromAndTo = l.getLookupKeysByFromAndTo(from, to, start, endKey, limit)
+	}
+	if len(ids) > 0 && from == "" && to == "" {
+		lookupKeys = lookupKeysByIDs
```
peterbroadhurst, line 130: This path doesn't seem to drop into the `results := l.getReceiptsByLookupKey(lookupKeys, limit)` branch (because of an `else if`) - so this result is un-used.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	} else if len(ids) == 0 && (from != "" || to != "") {
+		lookupKeys = lookupKeysByFromAndTo
```
peterbroadhurst, line 132: As above

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	} else if len(ids) > 0 && (from != "" || to != "") {
+		lookupKeys = intersect(lookupKeysByIDs, lookupKeysByFromAndTo)
+		sort.Sort(sort.Reverse(sort.StringSlice(lookupKeys)))
+		results := l.getReceiptsByLookupKey(lookupKeys, limit)
+		return results, nil
+	}
+
+	// no reference points, iterate normally
+	index := 0
```
peterbroadhurst, line 141: It might be more readable to have this as a separate `getReceiptsNoFilter` function that's called.

... I'll admit it does seem duplicative to have this, as I guess the only difference here is there's no filtering on the loop.
But I understand the loop with filtering is rather complex.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	sort.Sort(sort.StringSlice(result))
+	// since our query searches in descending order, while sort.SearchString requires ascending order
+	// the "start" is the end, and "end" is the start
+	result = applyBound(result, end, start)
+	return result
+}
+
+func (l *levelDBReceipts) getLookupKeysByFromAndTo(from, to string, start, end string, limit int) []string {
+	if from != "" || to != "" {
```
peterbroadhurst, line 241: Could avoid deep nesting by:
```go
if from == "" && to == "" {
  return nil
}
```

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+		fromKeys := []string{}
+		itr := l.store.NewIterator()
+		defer itr.Release()
+
+		if from != "" {
+			count := 0
```
peterbroadhurst, line 247: Seems like this block is a duplicate of the `to` block, with just the string different in the inserts, and `[]string` being a different set.
Could be split into a function to avoid the code duplication.

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+			return &results, nil
+		}
+	} else {
+		// create the iterator without a range
+		itr = l.store.NewIterator()
+	}
+	defer itr.Release()
+
+	if limit == 0 {
+		limit = l.defaultLimit
```
peterbroadhurst, line 115: Glad to see this included 👍

internal/rest/leveldbreceipts.go
```diff
@@ -0,0 +1,373 @@
...
+		return nil, errors.Errorf(errors.ReceiptStoreLevelDBConnect, err)
+	}
+	t := time.Unix(1000000, 0)
+	entropy := ulid.Monotonic(rand.New(rand.NewSource(t.UnixNano())), 0)
+
+	return &levelDBReceipts{
+		conf:         conf,
+		store:        store,
+		idEntropy:    entropy,
+		defaultLimit: 50,
```
peterbroadhurst, line 51: We should make this configurable

peterbroadhurst (commented):
This is a great piece of engineering.
Loads of tough problems you've worked through  💪

There are a few minor comments that need to be worked through before merge (mutex locking, on the config, and a couple of what seem like dead branches), and I've added some suggestions.

I don't think any of these would invalidate your end-to-end testing under load though.

internal/rest/leveldbreceipts.go
jimthematrix: good idea.

internal/rest/leveldbreceipts.go
jimthematrix: all 3 branches should be used:
- if `ids` are present, and both `from` and `to` are empty, branch 1
- if `ids` are empty, and at least one of `from` and `to` is non-empty, branch 2
- if `ids` are present, and one of `from` and `to` is non-empty, branch 3

peterbroadhurst: Agreed - I just mean in the code, there are two branches where all they do is set `lookupKeys`, but there's no reference to `lookupKeys` below that in the function.

peterbroadhurst:
Hey @jimthematrix - reviewed through and see there's still a number of comments open.
Wonder if you have a thought on the best way to close this one out, as I know you have other competing priorities in the Fabric space for FireFly. If you want me to help close out some of the open items, let me know.

peterbroadhurst:
Would also be good to have a summary of the function added, the design, and the externals config (once finalized) in the main comment on the PR for reference.

jimthematrix:
Thanks @peterbroadhurst.

Just added a summary of the design and supported functions, as well as the configurations.

Sorry I missed the option to disable the additional indexing entries, thus disabling query parameters (`since`, `from` and `to`). Can get to that in a bit or you are totally welcome to contribute a commit.

internal/rest/restgateway.go
peterbroadhurst: Retracting this comment, as I tracked this through the config and agree this is the most consistent 👍

internal/rest/leveldbreceipts.go
peterbroadhurst: Discussed with @jimthematrix and this is deferred. Performance testing did not show an issues in the millions of transaction range.

internal/rest/leveldbreceipts.go
peterbroadhurst: I added a commit and unit test to resolve this.

internal/rest/leveldbreceipts.go
peterbroadhurst: Thanks for adding this 👍

peterbroadhurst (approved):
Thanks @jimthematrix - went through in detail, to confirm we'd closed out all the pending conversations. Excited to see this go in 👍

PR #130, opened 2021-07-13
Catchup mode: read pages of blocks, when we're a long way behind

peterbroadhurst:
The ethereum log streaming interface, doesn't have any ability to limit/paginate calls to `eth_getFilterLogs` / `eth_getFilterChanges`, per https://github.com/ethereum/go-ethereum/issues/17487

This is problematic on chains with large numbers of blocks & transactions, when trying to reply events from a long way behind the current block.

Currently the code performs an `eth_getFilterLogs` from the latest checkpoint -> `latest`, which could be millions of blocks in the case of a new node joining the network.

So this PR proposes adding an automatically enabled "catchup" mode, in the case that a stream is restarting and finds itself a long way behind in the latest block on the node it's on. 
In this mode we repeat the following to read pages of blocks:
1. `eth_blockNumber`
2. `eth_getLogs` - checkpoint (or last page end), to end of page
... that happens until we find we're get within that window of the head again.

Default is to enable it when we are restarting `250` blocks or more behind the head, and read `250` blocks at a time.

There are some unrelated lint fixes in this PR too that were picked up by my compiler.

### DIFF
--- a/internal/events/submanager.go
+++ b/internal/events/submanager.go
@@ -42,6 +42,9 @@ const (
 	subIDPrefix        = "sb-"
 	streamIDPrefix     = "es-"
 	checkpointIDPrefix = "cp-"
+
+	defaultCatchupModeBlockGap = int64(250)
+	defaultCatchupModePageSize = int64(250)
 )
 
 // SubscriptionManager provides REST APIs for managing events
@@ -75,6 +78,8 @@ type subscriptionManager interface {
 type SubscriptionManagerConf struct {
 	EventLevelDBPath        string `json:"eventsDB"`
 	EventPollingIntervalSec uint64 `json:"eventPollingIntervalSec,omitempty"`
+	CatchupModeBlockGap     int64  `json:"catchupModeBlockGap,omitempty"`
+	CatchupModePageSize     int64  `json:"catchupModePageSize,omitempty"`
 	WebhooksAllowPrivateIPs bool   `json:"webhooksAllowPrivateIPs,omitempty"`
 }
 
@@ -107,6 +112,12 @@ func NewSubscriptionManager(conf *SubscriptionManagerConf, rpc eth.RPCClient, ws
 	}
 	if conf.EventPollingIntervalSec <= 0 {
 		conf.EventPollingIntervalSec = 1
+	}
+	if conf.CatchupModeBlockGap <= 0 {
+		conf.CatchupModeBlockGap = defaultCatchupModeBlockGap
+	}
+	if conf.CatchupModePageSize <= 0 {
+		conf.CatchupModePageSize = defaultCatchupModePageSize
 	}
 	return sm
 }
--- a/internal/events/subscription.go
+++ b/internal/events/subscription.go
@@ -86,7 +86,7 @@ func newSubscription(sm subscriptionManager, rpc eth.RPCClient, addr *ethbinding
 		logName:             i.ID + ":" + ethbind.API.ABIEventSignature(event),
 		filterStale:         true,
 		catchupModeBlockGap: sm.config().CatchupModeBlockGap,
-		catchupModePageSize: sm.config().CatchupModeBlockGap,
+		catchupModePageSize: sm.config().CatchupModePageSize,
 	}
 	f := &i.Filter
 	addrStr := "*"
--- a/internal/events/subscription.go
+++ b/internal/events/subscription.go
@@ -56,15 +56,18 @@ type SubscriptionInfo struct {
 
 // subscription is the runtime that manages the subscription
 type subscription struct {
-	info           *SubscriptionInfo
-	rpc            eth.RPCClient
-	lp             *logProcessor
-	logName        string
-	filterID       ethbinding.HexBigInt
-	filteredOnce   bool
-	filterStale    bool
-	deleting       bool
-	resetRequested bool
+	info                *SubscriptionInfo
+	rpc                 eth.RPCClient
+	lp                  *logProcessor
+	logName             string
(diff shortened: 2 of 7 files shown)


peterbroadhurst:
Example output from catchup mode on a new subscription starting from block zero:
```
[2021-07-13T14:12:02.892-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 0 -> 249
[2021-07-13T14:12:13.054-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 250 -> 499
[2021-07-13T14:12:23.147-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 500 -> 749
[2021-07-13T14:12:33.236-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 750 -> 999
[2021-07-13T14:12:43.409-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 1000 -> 1249
[2021-07-13T14:12:53.501-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 1250 -> 1499
[2021-07-13T14:13:03.613-04:00]  INFO sb-6e918cc8-7f2b-484d-4004-98a596f3b68b:Changed(address,int64,string,string): catchup mode. Blocks 1500 -> 1749
```

internal/events/submanager.go
```diff
@@ -108,6 +113,12 @@ func NewSubscriptionManager(conf *SubscriptionManagerConf, rpc eth.RPCClient, ws
 	if conf.EventPollingIntervalSec <= 0 {
 		conf.EventPollingIntervalSec = 1
 	}
+	if conf.CatchupModeBlockGap <= 0 {
```
vdamle, line 116: Nit.. Could we change the type of `CatchupModeBlockGap` to `uint` so there's no need to check for negative values?

internal/events/subscription.go
```diff
@@ -77,11 +80,13 @@ func newSubscription(sm subscriptionManager, rpc eth.RPCClient, addr *ethbinding
...
-		lp:          newLogProcessor(i.ID, event, stream),
-		logName:     i.ID + ":" + ethbind.API.ABIEventSignature(event),
-		filterStale: true,
+		info:                i,
+		rpc:                 rpc,
+		lp:                  newLogProcessor(i.ID, event, stream),
+		logName:             i.ID + ":" + ethbind.API.ABIEventSignature(event),
+		filterStale:         true,
+		catchupModeBlockGap: sm.config().CatchupModeBlockGap,
+		catchupModePageSize: sm.config().CatchupModeBlockGap,
```
vdamle, line 89:
```suggestion
		catchupModePageSize: sm.config(). CatchupModePageSize,
```

peterbroadhurst:
Example of it exiting catchup, and going back to normal operation:

```
[2021-07-13T14:16:57.024-04:00]  INFO sb-f839d72b-b690-4569-5eb3-efcbf01461f7:Changed(address,int64,string,string): catchup mode. Blocks 321000 -> 321249
[2021-07-13T14:17:07.395-04:00] DEBUG sb-f839d72b-b690-4569-5eb3-efcbf01461f7:Changed(address,int64,string,string): Marking filter stale=false, current sub filter stale=true
[2021-07-13T14:17:07.396-04:00]  INFO sb-f839d72b-b690-4569-5eb3-efcbf01461f7:Changed(address,int64,string,string): created filter from block 321250: 0x884e3b1e8e4718d0de26bc38576e1c15 - {Addresses:[] Topics:[[0x9cc76f31dd4253fe222c0bcccf929a17496fc4f64167bf8cd1910b9e16f5be0f]]}
```

internal/events/subscription.go
```diff
@@ -168,12 +173,39 @@ func (s *subscription) restartFilter(ctx context.Context, since *big.Int) error
...
+	}
+
+	blockNumber := ethbinding.HexBigInt{}
+	err := s.rpc.CallContext(ctx, &blockNumber, "eth_blockNumber")
+	if err != nil {
+		return errors.Errorf(errors.RPCCallReturnedError, "eth_blockNumber", err)
+	}
+
+	blockGap := new(big.Int).Sub(blockNumber.ToInt(), since).Int64()
+	if s.catchupModeBlockGap > 0 && blockGap > s.catchupModeBlockGap {
```
vdamle, line 201: Just a note that this will be a no-op and we won't try to re-catch up when a node is being re-synced or if there's an accidental chain rewind to a historical block with state after a non-graceful shutdown of the node. Think that is the right behavior given there's a stored checkpoint that tells us those blocks are already processed, maybe worth adding a comment.

vdamle (approved):

internal/events/submanager.go
peterbroadhurst: Thanks @vdamle - when I went through and tried making all the changes for this, I found the downstream complexity of casting and dealing with mapping to `big.Int` with comparison, became much more complicated than leaving it as an `int64`. So if it's ok, I'm going to leave this one as is.

internal/events/subscription.go
peterbroadhurst: Thanks for catching this one 👍

PR #131, opened 2021-07-14
Fix catchup for restored subscriptions, and add log

peterbroadhurst:
#130 missed the configuration of block catchup for restored subscriptions


vdamle (approved):
🏃🏼

PR #136, opened 2021-08-05
Handle spurious acks

peterbroadhurst:
Currently there are two cases where the ack processing can get stuck:
- An application sends a spurious additional ack
  - The WS code tries to deliver this to the eventstream code, while the ES code is trying to wait for a message
  - Eventually results in `We were not available to process it after ... seconds` in the logs
- The ES code receives an update to the eventstream, while it's waiting for an ack
  - This results in the WS code never being able to deliver the ack, so never delivering the next message

The proposed change in this PR adds a buffer of 1 to the channel between the WS and ES code, and changes the timeout into a `default` when delivering the ack. This prevents the deadlock possibility. There are always either `0` or `1` acks in the pipe.

### DIFF
--- a/internal/ws/wsconn.go
+++ b/internal/ws/wsconn.go
@@ -163,7 +163,7 @@ func (c *webSocketConnection) handleAckOrError(t *webSocketTopic, err error) {
 		log.Debugf("WS/%s: response (error='%t') on topic '%s' passed on for processing", c.id, isError, t.topic)
 		break
 	default:
-		log.Debugf("WS/%s: suprious ack received (error='%t') on topic '%s'", c.id, isError, t.topic)
+		log.Debugf("WS/%s: spurious ack received (error='%t') on topic '%s'", c.id, isError, t.topic)
 		break
 	}
 }
--- a/internal/ws/wsconn.go
+++ b/internal/ws/wsconn.go
@@ -18,7 +18,6 @@ import (
 	"reflect"
 	"strings"
 	"sync"
-	"time"
 
 	ws "github.com/gorilla/websocket"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


internal/ws/wsserver_test.go
```diff
@@ -192,16 +192,27 @@ func TestTimeoutProcessing(t *testing.T) {
...
+		Topic: "mytopic",
+	})
+	c.WriteJSON(&webSocketCommandMessage{
+		Type:  "ack",
+		Topic: "mytopic",
 	})
+	c.Close()
 
-	// Confirm we close the connection after the timeout pops
 	for len(w.connections) > 0 {
```
awrichar, line 204: This loop can be dropped now, right?

internal/ws/wsserver_test.go
```diff
@@ -192,16 +192,27 @@ func TestTimeoutProcessing(t *testing.T) {
...
 		time.Sleep(1 * time.Millisecond)
 	}
 
+	for _, conn := range w.connections {
```
awrichar, line 208: I'm also confused about whether `w.connections` is populated at this point. Seems like after the above loop it would be empty, so this would be a no-op? Maybe I'm not following correctly.

internal/ws/wsconn.go
```diff
@@ -160,11 +159,11 @@ func (c *webSocketConnection) listen() {
...
 	isError := err != nil
 	select {
-	case <-time.After(c.server.processingTimeout):
-		log.Errorf("WS/%s: response (error='%t') on topic '%s'. We were not available to process it after %.2f seconds. Closing connection", c.id, isError, t.topic, c.server.processingTimeout.Seconds())
-		c.close()
 	case t.receiverChannel <- err:
 		log.Debugf("WS/%s: response (error='%t') on topic '%s' passed on for processing", c.id, isError, t.topic)
 		break
+	default:
+		log.Debugf("WS/%s: suprious ack received (error='%t') on topic '%s'", c.id, isError, t.topic)
```
awrichar, line 166: typo: spurious

awrichar:
Tried this with my local FireFly stack and it seems to correctly detect spurious acks. Can't tell if the unit test is actually testing it as desired, but it's possible I'm not following the test entirely.

internal/ws/wsserver_test.go
awrichar: Or maybe you are waiting for w to close after c is closed...

peterbroadhurst: Agreed - it's actually the second loop that needs to be dropped though 👍

internal/ws/wsserver_test.go
peterbroadhurst: resolved by above

awrichar (approved):

vdamle (approved):

PR #141, opened 2021-09-01
Add fly-transaction param to query transaction details

awrichar:
Performing a GET to a method endpoint with fly-transaction set to a previously
executed transaction hash will return details on that transaction, including
the decoded input arguments.

~~I still need to add test coverage, but~~ I wanted to vet that this is a good design.
In particular want to see if we definitely want this on the method REST endpoint,
or if it should be at a higher level somehow.

~~Note that this branch also includes the enhancement from #140~~

### DIFF
--- a/internal/eth/txn.go
+++ b/internal/eth/txn.go
@@ -69,11 +69,11 @@ type TxnInfo struct {
 	BlockNumber      *ethbinding.HexBigInt `json:"blockNumber,omitempty"`
 	From             *ethbinding.Address   `json:"from,omitempty"`
 	To               *ethbinding.Address   `json:"to,omitempty"`
-	Gas              *ethbinding.HexBigInt `json:"gas"`
+	Gas              *ethbinding.HexUint64 `json:"gas"`
 	GasPrice         *ethbinding.HexBigInt `json:"gasPrice"`
 	Hash             *ethbinding.Hash      `json:"hash"`
-	Nonce            *ethbinding.HexUint   `json:"nonce"`
-	TransactionIndex *ethbinding.HexUint   `json:"transactionIndex"`
+	Nonce            *ethbinding.HexUint64 `json:"nonce"`
+	TransactionIndex *ethbinding.HexUint64 `json:"transactionIndex"`
 	Value            *ethbinding.HexBigInt `json:"value"`
 	Input            *ethbinding.HexBytes  `json:"input"`
 }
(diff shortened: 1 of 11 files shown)


awrichar:
Postman

awrichar:
Error if an invalid hash is provided:

Postman

internal/eth/txn.go
```diff
@@ -63,6 +63,21 @@ type TxnReceipt struct {
...
+	BlockNumber      *ethbinding.HexBigInt `json:"blockNumber,omitempty"`
+	From             *ethbinding.Address   `json:"from,omitempty"`
+	To               *ethbinding.Address   `json:"to,omitempty"`
+	Gas              *ethbinding.HexBigInt `json:"gas"`
+	GasPrice         *ethbinding.HexBigInt `json:"gasPrice"`
+	Hash             *ethbinding.Hash      `json:"hash"`
+	Nonce            *ethbinding.HexUint   `json:"nonce"`
+	TransactionIndex *ethbinding.HexUint   `json:"transactionIndex"`
+	Value            *ethbinding.HexBigInt `json:"value"`
+	Input            *ethbinding.HexBytes  `json:"input"`
```
awrichar, line 78: How do I know the types for these values returned by geth? Particularly on some of the BigInt vs Uint I just guessed... should they all just be BigInt?

peterbroadhurst: The official reference is here: https://eth.wiki/json-rpc/API#eth_gettransactionbyhash
However, I find going straight to the code best:
https://github.com/ethereum/go-ethereum/blob/9ada4a2e2c415e6b0b51c50e901336872e028872/internal/ethapi/api.go#L1264-L1284

One quick conversation to have here before we merge, is that ethconnect does try and make various returns more friendly in other places. Giving values in non-hex decimal strings, as well as hex.
I'm wondering if we should be consistent here with the mapping.

See: https://github.com/hyperledger-labs/firefly-ethconnect/blob/984cba49f715ea3f5535261d4b38564816065a0a/internal/messages/messages.go#L141-L163

peterbroadhurst (commented):
One comment for consideration.

internal/eth/txn.go
awrichar: Yea the docs just listed everything as `QUANTITY` which was why I couldn't tell the expected sizes. Thanks for the code pointer! Looks like I got pretty close, except that `Gas`, `Nonce`, and `TransactionIndex` can all be `Uint64`. I'll update.

I guess it makes sense to be consistent with how the fields are formatted between this and the `TransactionReceipt` you linked as well.

awrichar:
Not entirely sure what's going on with the codecov bot here either. Clicking on the coverage report seems to indicate a slight _increase_ in coverage, but the comment here on the PR indicates a slight _decrease_. Maybe I'm reading wrong?

awrichar:
New JSON body with number strings and hex strings.
Postman

peterbroadhurst (approved):
Thanks for doing the consistency work on the return here 👍

PR #144, opened 2021-09-09
Split "contracts" package into "contractgateway" and "contractregistry"

awrichar:
The new "contractregistry" package has minimal dependencies and can be used from anywhere
to lookup contracts/ABIs, while the "contractgateway" package remains large and pulls in nearly
everything.

This will facilitate future enhancements where we want to do ABI lookups from outside the gateway
(such as from the "events" package).


peterbroadhurst (approved):
Thanks for factoring this out! A great step forwards

PR #146, opened 2021-09-10
Move contract sorting up to contractStore

awrichar:
Makes the contractStore unit tests pass more reliably.


internal/contractgateway/smartcontractgw.go
```diff
@@ -337,7 +337,6 @@ func (g *smartContractGW) writeAbiInfo(requestID string, msg *messages.DeployCon
 	return nil
 }
 
-// listContracts sorts by Title then Address and returns an array
```
peterbroadhurst, line 340:
```suggestion
// listContractsOrABIs sorts by Title then Address and returns an array
```

peterbroadhurst (approved):
One comment change request, with approval

PR #147, opened 2021-09-13
Allow including method inputs with subscribed events

awrichar:
Passing {"inputs": true} when initializing an event stream will cause each
event to include two new fields - inputMethod will be the name of the method
that emitted the event, and inputArgs will be a map of the args passed in
when the method was called.

This can be reviewed in parallel with #141, although they share one commit
containing some generic helpers for decoding transaction info.


internal/events/logprocessor.go
```diff
@@ -28,13 +28,15 @@ import (
...
-	Timestamp        uint64               `json:"timestamp,omitempty"`
+	Address          ethbinding.Address     `json:"address"`
+	BlockNumber      ethbinding.HexBigInt   `json:"blockNumber"`
+	TransactionIndex ethbinding.HexUint     `json:"transactionIndex"`
+	TransactionHash  ethbinding.Hash        `json:"transactionHash"`
+	Data             string                 `json:"data"`
+	Topics           []*ethbinding.Hash     `json:"topics"`
+	Timestamp        uint64                 `json:"timestamp,omitempty"`
+	InputMethod      string                 `json:"inputMethod,omitempty"`
+	InputArgs        map[string]interface{} `json:"inputArgs,omitempty"`
```
awrichar, line 39: I couldn't entirely figure out the reason for the existence of both `eventData` and `logEntry`. The two structs seem almost identical and most data is just copied from one to the other, which seems less efficient than it could be.

internal/events/subscription.go
```diff
@@ -237,6 +253,29 @@ func (s *subscription) getEventTimestamp(ctx context.Context, l *logEntry) {
 	s.lp.stream.blockTimestampCache.Add(blockNumber, l.Timestamp)
 }
 
+func (s *subscription) getTransactionInputs(ctx context.Context, l *logEntry) {
+	abi, err := loadABI(s.cr, s.info.ABI)
```
awrichar, line 257: This is obviously pretty heavyweight - for every event, we fetch the ABI and then iterate over all the methods to determine which one was called. Would appreciate thoughts on the best ways and places to cache things. Perhaps a method cache on `ContractResolver`?

internal/contractregistry/contractstore.go
```diff
@@ -161,22 +175,42 @@ func (cs *contractStore) GetContractByAddress(addrHex string) (*ContractInfo, er
 	return info.(*ContractInfo), nil
 }
 
-func (cs *contractStore) GetABIByID(id string) (*messages.DeployContract, *ABIInfo, error) {
+func (cs *contractStore) GetABI(location *ABILocation, refresh bool) (*messages.DeployContract, string, error) {
+	switch location.ABIType {
```
awrichar, line 179: Is there any better way to generically store a reference to an ABI? I could only pare down to these 3 different ways of retrieving an ABI, so I was left with this "ABILocation" that pairs an enum with a string identifier...

internal/events/subscription.go
peterbroadhurst: Yeah, I think a method ID based cache on ContractResolver is important. I agree this could be very heavyweight without that.

There's a nice and flexible LRU cache facility used in FireFly core here:
https://github.com/hyperledger-labs/firefly/blob/c9d070f5f5b9ee07842fb1cc1cb8baec54231f9e/internal/privatemessaging/groupmanager.go#L152-L156

peterbroadhurst (changes requested):
Sorry to add work, but I do think we want to pop the cache in.
One other comment on whether we need to update `logEntry` too (as that should I think be just the back-side JSON model).

internal/events/logprocessor.go
awrichar: @peterbroadhurst did you have any info on this distinction? Currently I'm performing the transaction lookup in `subscription.go` while building the `logEntry`, and everything is copied from there to the `eventData`.

One other option would be to not put any of it on `logEntry`, and to only do the parsing when building `eventData` - this would require passing the `ContractResolver` and `ABILocation` from the subscription down into `logprocessor.go`.

internal/events/subscription.go
awrichar: That looks to be using https://github.com/karlseguin/ccache
Here in ethconnect, eventstream is already using https://github.com/hashicorp/golang-lru

Seems like we'd only want one per project, so I'll probably start with the hashicorp one unless there's a good reason to switch.

awrichar:
Added a first pass on the LRU cache. Might still need some tweaking, and test coverage is kind of blocked on #148 - but it's close.

internal/events/subscription.go
peterbroadhurst: Sounds good. My recollection for the reason we used `ccache` in FireFly, was that `golang-lru` didn't support sizing the objects (in Kb) - only a count.

Here only a count seems fine, so agree with sticking for consistency.

internal/contractregistry/contractstore.go
```diff
@@ -58,17 +65,22 @@ type contractStore struct {
...
+func NewContractStore(baseURL, storagePath string, rr RemoteRegistry) (cs ContractStore, err error) {
+	c := &contractStore{
 		baseURL:               baseURL,
 		storagePath:           storagePath,
 		rr:                    rr,
 		contractIndex:         make(map[string]messages.TimeSortable),
 		contractRegistrations: make(map[string]*ContractInfo),
 		abiIndex:              make(map[string]messages.TimeSortable),
 	}
+	if c.abiCache, err = lru.New(DefaultABICacheSize); err != nil {
```
peterbroadhurst, line 80: Would it make sense to add a config option for that?

I see the caller has to pass in config explicitly, with `storagePath` being an input. Might be worth having a `ContractStoreConfig` object, similar to other structs.

Not a strong objection here on this one to starting with a `const`, so will leave it with you whether to do this before merge.

peterbroadhurst (approved):
Just one extra comment to consider before merging @awrichar

PR #148, opened 2021-09-14
Use mockery for contractregistry and sarama mocks

awrichar:
This makes mocks a lot clearer and more explicit about the expected calls.


Makefile
```diff
@@ -4,6 +4,11 @@ BINARY_UNIX=$(BINARY_NAME)-tux
 BINARY_MAC=$(BINARY_NAME)-mac
 BINARY_WIN=$(BINARY_NAME)-win
 
+GOBIN := $(shell $(VGO) env GOPATH)/bin
+MOCKERY := $(GOBIN)/mockery
+
+SARAMA_PATH := $(shell $(VGO) list -f '{{.Dir}}' github.com/Shopify/sarama)
```
peterbroadhurst, line 10: I'm not sure having this at the head of the file will work, as it would happen on an empty env before `go get`.

Here's an example of how to do a command like this within a task, if that helps:
https://github.com/hyperledger-labs/firefly/blob/32d09ec27ca3580407a3781e1c2e2f752804f0cb/Makefile#L20

peterbroadhurst (approved):
Thanks for making the time for this, great to have consistency here within the FireFly stable.
One small item to think about, before merging.

Makefile
awrichar: Might also be able to just replace `:=` with `=` - I'll try around on a fresh env.

awrichar: Guess that might result in a lot of `gets` though. Will see what I can come up with...

awrichar: So after testing, this actually had the opposite problem to what I expected. `go list` implies downloading the named dependencies if they aren't downloaded already, so with a clean Go environment you'd actually have a very long first invocation of `make`, since it would evaluate this variable and download sarama before even processing the rest of the Makefile.

Added a commit to set the variable only when actually ready to build mocks. Thanks for the catch!

PR #154, opened 2021-09-29
Added support for inline schema of transactions and events payload

jimthematrix:
This allows ethconnect to be used with contracts that it doesn't have ABIs for, by passing the schema for the transaction payload or event inline.

### Sending Transactions
In the `POST /contracts/:address/:method` requests:
```json
{
    "$schema":  {
      "type": "function",
      "name": "put",
      "stateMutability": "nonpayable",
      "inputs": [
        {
          "name": "x",
          "type": "uint256",
          "internalType": "uint256"
        }
      ],
      "outputs": []
    },
    "x": "12346"
}
```

### Subscribing to Events
In the `POST /contracts/:address/:event/subscribe` requests:
```json
{
    "$schema": {
      "anonymous": false,
      "inputs": [
        {
          "internalType": "uint256",
          "name": "data",
          "type": "uint256"
        }
      ],
      "name": "DataStore",
      "type": "event"
    },
    "stream": "es-9655024d-85cc-4f56-7773-ab6d96fc861e"
}
```

In the above examples, the ethconnect instance does not need to know anything about the contract at `:address` either in the remote registry or in the local ABI store.

### DIFF
--- a/internal/contractgateway/rest2eth.go
+++ b/internal/contractgateway/rest2eth.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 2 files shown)


peterbroadhurst:
@jimthematrix - I think the key question for me here, is why this isn't the existing payload we have already?
The introduction of a brand new, but overlapping, route under `/contracts` seems to be a little awkward.

e.g. why isn't the transaction submission payload same JSON payload I can send on `POST` to `/`, or more importantly put directly onto the Kafka input topic.

I note the ability to stream transactions directly in over Kafka was brought up on the community call today.

I'm going to try and draw a diagram of my understanding of the overall job at hand here around on-chain logic (assuming that's not somewhere for em to refer to @nguyer).

peterbroadhurst:
The transaction submission payload we have today for the "webhook" or Kafka input is here:
https://github.com/hyperledger/firefly-ethconnect/blob/e708f38f5eac8ad2c4a6b084577c3954e75e7488/internal/messages/messages.go#L117-L122

jimthematrix:
@peterbroadhurst  I'm aware of the payload available to the POST calls to `/`, `/hook` and `/fasthook` endpoints. But I thought those endpoints were being deprecated in favor of the more "RESTful" endpoints with `/contracts/:address/:method` and `/contracts/:address/:method/:subcommand` which accomplish the same thing. maybe I'm confused by why we have these endpoints at the same time.

peterbroadhurst:
Here's a link to the issue in FireFly core where I've tried to get the architecture discussion up and running:
https://github.com/hyperledger/firefly/issues/217#issuecomment-930482302

peterbroadhurst:
> But I thought those endpoints were being deprecated in favor of the more "RESTful" endpoints with

That might have been true in previous uses of ethconnect. But we are moving in precisely the _opposite_ direction here for FireFly, in deprecating the RESTful endpoints.

I've tried to write up the fuller state here:
https://github.com/hyperledger/firefly/issues/217#issuecomment-930488413

peterbroadhurst:
Closing this, as I do not believe the outcome will be any code changes (per above discussion).

PR #158, opened 2021-10-07
Honor fly-id for contract deployments

awrichar:
This also prevents bugs where the loaded deployMsg already had an ID and
the ID is reused.

Signed-off-by: Andrew Richardson <andrew.richardson@kaleido.io>


vdamle:
Tested a branch build that multiple instances of contract can be deployed with this fix:

Issue #159, opened 2021-10-08
FireFly e2e test failing with FF10111: Error from ethconnect "No ABI found with ID"


peterbroadhurst:
```
[34mfirefly_core_0_1  |[0m Error: FF10111: Error from ethconnect: {&#34;error&#34;:&#34;No ABI found with ID &#34;}
[34mfirefly_core_0_1  |[0m Usage:
[34mfirefly_core_0_1  |[0m   firefly [flags]
[34mfirefly_core_0_1  |[0m   firefly [command]
[34mfirefly_core_0_1  |[0m 
[34mfirefly_core_0_1  |[0m Available Commands:
[34mfirefly_core_0_1  |[0m   help        Help about any command
[34mfirefly_core_0_1  |[0m   showconfig  List out the configuration options
[34mfirefly_core_0_1  |[0m 
[34mfirefly_core_0_1  |[0m Flags:
[34mfirefly_core_0_1  |[0m   -f, --config string   config file
[34mfirefly_core_0_1  |[0m   -h, --help            help for firefly
[34mfirefly_core_0_1  |[0m 
[34mfirefly_core_0_1  |[0m Use "firefly [command] --help" for more information about a command.
[34mfirefly_core_0_1  |[0m 
[34mfirefly_core_0_1  |[0m FF10111: Error from ethconnect: {&#34;error&#34;:&#34;No ABI found with ID &#34;}
```

peterbroadhurst:
Looks like the code that ensures an ID is set on the header, might only exist in the Kafka codepath:
https://github.com/hyperledger/firefly-ethconnect/blob/8545f1c09d5fc9829508dac9a08d8c57210079eb/internal/kafka/kafkabridge.go#L172-L174

peterbroadhurst:
Hmmn, looks like back then it was still only set on the `webhooks` path, not the `resttoeth` path.
So the problem might always have been there for this path, but masked by the bug just fixed in https://github.com/hyperledger/firefly-ethconnect/pull/158

PR #139, opened 2021-08-13
Allow client to specify request ID

awrichar:

PR #160, opened 2021-10-08
Assign message ID before dispatching message to Kafka in all cases

peterbroadhurst:
Currently an ID is only assigned to the message in these cases:
- When using the Webhooks route to input a raw JSON message (`/` or `/hooks`) - before dispatching (to Kafka/sync)
- When explicitly passing a `fly-id`/`kaleido-id` param to the REST API Gateway - before dispatching (to Kafka/sync)
  - Added in #139 
- When the message is received at the Kafka side, and we find no ID inside there

This means in the Kakfa case there is always a message ID assigned at the point of processing.
But in the Sync dispatcher, as used by the FireFly CLI environment, there was no ID being assigned.

We hit an issue with #139 where we found it was leaking through an ID in the REST API Gateway case, that actually came from the gateway ID - that was just fixed in #158.

That fix then in turn exposed the problem of messages ended up with _no ID at all_ and cause the break in the e2e builds for FireFly seen in #159


internal/contractgateway/smartcontractgw.go
```diff
@@ -216,7 +216,12 @@ func (g *smartContractGW) PostDeploy(msg *messages.TransactionReceipt) error {
 				err = g.cs.AddRemoteInstance(msg.RegisterAs, "0x"+addrHexNo0x)
 			}
 		} else {
-			_, err = g.cs.AddContract(addrHexNo0x, requestID, registeredName, msg.RegisterAs)
+			abiID := requestID
+			if msg.Headers.ReqABIID != "" {
```
awrichar, line 220: So there's still a case where we expect this header to be missing, and we fall back to using the requestID?

vdamle:
> As an aside, I would propose that @awrichar be considered for promoting to a core contributor, due to his significant recent contributions to Ethconnect.

Agree that @awrichar can be promoted to core contributor 👍🏼

PR #166, opened 2021-10-11
Update ref to latest ethbinding

peterbroadhurst:


peterbroadhurst:
I could not see the reason for the test failure, which looks to me like a timing condition where a background processing of one of our `eventstreams` tests tries to fire the `GetABI` mock after the test has finished.

So I've added `-v` to the `go test` runner, just for when it's running in a Github action (not locally with `make` as that would be really annoying).

vdamle (approved):

Issue #173, opened 2021-11-17
Provide a circuit-breaker in the case that the consumer falls excessively behind producers


peterbroadhurst:
Kafka partitions have a limited buffer, and a well maintained Kafka environment will have controls over the size of that buffer - it will not be infinite (limited by time and/or size).

Unlike traditional queuing systems, Kafka does not have the concept of a "queue full" condition, or even the concept of consuming messages. It simply tracks the offsets that different consumers have seen on a continual stream of messages.

This means that if your consumers fall too far behind your producers, and a size-based log is configured, then Kafka will discard messages in-flight.

In Ethconnect there's currently no feedback when this happens. The messages are just lost. This is problematic for applications that are relying on Ethconnect to reliably stream all their transactions in - fire&forget style. If they overrun the input of messages, they can end up with missing receipts.

So this issue tracks a new feature enhancement, to provide an option that causes the _sending_ side of Ethconnect, to have a background task running to monitor the gap between the messages being sent, and the offset being consumed.

The logic proposed is pretty simple:
- Before sending a message, after the payload has been constructed, we call the circuit breaker:
  - We provide the size of the `[]byte` that will be sent
- The circuit breaker keeps track of the average size of all these requests
- The circuit breaker regularly polls Kafka to check how things are looking (configurable interval - 1 min etc.)
  - Uses the consumer interface to query partitions and high water marks:
      - https://pkg.go.dev/github.com/Shopify/sarama#NewConsumerFromClient
  - Uses the offset manager interface to query consumer group offsets:
      - https://pkg.go.dev/github.com/Shopify/sarama#NewOffsetManagerFromClient
  - Requires the safe buffer size to be passed in as a configuration parameter
      - Complex to query this, and should not be the full size because there's estimation involved
      - Example would be if a 100MB Kafka retention is configured, then the config would be set with a 75MB safe buffer
  - Uses the size estimation combined with the gap between the HWM and NextOffset for the consumer group
      - Estimates the amount of data in the buffer
  - Always logs this, to give a running commentary in the logs of the offset gap
  - Closes the circuit breaker if there's an excessive gap

peterbroadhurst:
After a bunch of prototyping, it seems like the route of having a separate consumer using the above approach has a lot of both code and runtime overhead.

The circuit breaker would have to have its own Consumer on each partition, to be able to query the high water marks. This means both code complexity, and an extra burden on the Kafka brokers.

[...]

PR #175, opened 2021-11-21
Add circuit breaker to stop runaway producers losing messages, and immediate receipt option

peterbroadhurst:
Implementations for:
- [x] #173 
- [x] #176

## Kafka Circuit Breaker

When enabled, the difference between the `offset` of the consumer and the current Kafka `HighWaterMark` of each partition being consumed within the runtime is monitored by a new runtime component

A running average of the size of the consumed messages is kept, to allow an estimation of the byte size of the gap between the offset and high water mark. This is needed as Kafka will roll the buffer based on the size (not the count) of messages, so we need to estimate that.

On startup a log is output showing the configuration.

```
[2021-11-22T04:02:56.465Z]  INFO CircuitBreakerEnabled: trip=6144.00Kb reset=4915.20Kb
```

Under normal consumption, new log entries are logged at most every X seconds (30 by default):

```
[2021-11-22T03:49:36.747Z]  INFO CircuitBreakerHealth: topic=zzzixar21f-zzlcxv8ksq-requests partition=0 offset=45233 hwm=45234 gap=1 gap.estimate=5.73Kb tripped=false lastTripped=never
```

In the case that the circuit breaker trips a log is output as follows - note that it shows the gap in messages, as well as the estimated size of the buffer represented by that gap. The switch trips when this meets an upper threshold.

```
[2021-11-22T01:08:24.551Z]  INFO CircuitBreakerTripped: topic=zzzixar21f-zzlcxv8ksq-requests partition=0 offset=3730 hwm=22513 gap=18783 gap.estimate=8492.70Kb tripped=true lastTripped=2021-11-22T01:08:24.551313415Z
```

When the circuit breaker un-trips (resets), the log output is as follows. This occurs when the gap estimate drops below a fraction of the upper bound - `0.8` (80%) by default:
```
[2021-11-22T01:09:45.179Z]  INFO CircuitBreakerReset: topic=zzzixar21f-zzlcxv8ksq-requests partition=0 offset=11619 hwm=22513 gap=10894 gap.estimate=4915.07Kb tripped=false lastTripped=2021-11-22T01:08:24.551313415Z
```

While tripped, any REST API requests attempting to send messages will receive a `500` error with the following body:

```json
{"error":"Unable to send Kafka message as the gap of 29507 messages between consumer and producer is too large. Estimated at 13312.73Kb"}
```

## Immediate Receipts

A new query parameter on REST API Gateway invocations of contract methods is provided `fly-acktype=receipts` (or `kld-acktype` in Kaleido environments with a custom prefix).

For raw webhooks, this can be supplied as an `ackmode: "receipts"` body parameter.

When specified in async requests, it will be immediately possible to query a receipt object on `/replies/:id`.

- `pending` will be set to `true` (not present in the final receipt once it replaces this entry)
- `msgAck` will be a string containing the Kafka message details
- `headers.type` will be the type of request `SendTransaction` or `DeployContract`
- The rest of the message that has been dispatched to Kafka will be included, with all inputs etc.

Example:

```json
{
  "_id": "97fd6f37-11e9-45d5-4355-4abc77279fef",
  "from": "0x4d1c5d7c4696afd597b2906564ea3aa9e0c382a6",
  "gas": 0,
  "gasPrice": 0,
  "headers": {
    "id": "97fd6f37-11e9-45d5-4355-4abc77279fef",
    "type": "SendTransaction"
  },
  "method": {
    "inputs": [
      {
        "internalType": "int64",
        "name": "someNumber",
        "type": "int64"
      },
      {
        "internalType": "string",
        "name": "someMessage",
        "type": "string"
      }
    ],
    "name": "set",
    "outputs": [],
    "stateMutability": "nonpayable",
    "type": "function"
  },
  "msgAck": "zzzixar21f-zzlcxv8ksq-requests:0:45233",
  "params": [
    "12345",
    "test"
  ],
  "pending": true,
  "receivedAt": 1637552976745,
  "to": "0xbb99253e723891b73b9c18bcc818678f3e29c009",
  "value": 0
}
```

Once the transaction is sent to the blockchain and succeeds/fails, this is replaced with the usual receipt - which is unchanged:

```json
{
  "_id": "97fd6f37-11e9-45d5-4355-4abc77279fef",
  "blockHash": "0x9df474173e82d41ddf03f7db4c2a605a944e0d30b38a326af4588eef0c1806da",
  "blockNumber": "3566",
  "cumulativeGasUsed": "39428",
  "from": "0x4d1c5d7c4696afd597b2906564ea3aa9e0c382a6",
  "gasUsed": "39428",
  "headers": {
    "id": "1faa4e96-4d2c-4ada-58a6-1eb9a51cc379",
    "requestABIId": "",
    "requestId": "97fd6f37-11e9-45d5-4355-4abc77279fef",
    "requestOffset": "zzzixar21f-zzlcxv8ksq-requests:0:45233",
    "timeElapsed": 5.369752082,
    "timeReceived": "2021-11-22T03:49:36.747344301Z",
    "type": "TransactionSuccess"
  },
  "nonce": "45284",
  "receivedAt": 1637552982120,
  "status": "1",
  "to": "0xbb99253e723891b73b9c18bcc818678f3e29c009",
  "transactionHash": "0x110cbf40c547fe5575c4b834115e89437dd335a98536bf869bce726005ae70ef",
  "transactionIndex": "0"
}
```

## Sample App

In case it's useful, here is a simple sample app for streaming messages in at a high pace into a runtime:

```js
const axios = require('axios');
const https = require('https');

const BASE_URL = 'https://zzzixar21f-zzlcxv8ksq-connect.dev2.photic.io/';
const USERNAME = 'xxxxxx';
const PASSWORD = 'xxxxxx;
const INSTANCE_ADDRESS = '0xbb99253e723891b73b9c18bcc818678f3e29c009';
const FROM_ADDRESS = '0x4d1c5d7c4696afd597b2906564ea3aa9e0c382a6';

const MAX_SOCKETS = 50;
const MAX_PROMISES = 200;
const TOTAL_MESSAGES = 10000;

const client = axios.create({
  httpsAgent: new https.Agent({
    maxSockets: MAX_SOCKETS,
  }),
  auth: {
    username: USERNAME,
    password: PASSWORD,
  },
  baseURL: BASE_URL
});

async function run() {

  let i = 0;
  const inflight = {};
  while (i < TOTAL_MESSAGES) {
    if (Object.keys(inflight).length < MAX_PROMISES) {
      const reqNo = ++i;
      inflight[`${reqNo}`] = (client.request({
        url: `instances/${INSTANCE_ADDRESS}/set`,
        method: 'POST',
        data: {
          "someMessage": `Test ${reqNo}`,
          "someNumber": reqNo,
        },
        params: {
          'kld-from': FROM_ADDRESS,
          'kld-sync': 'false',
          'kld-acktype': 'receipt',
        }
      })
        .then(({status})  => console.log(`PASS - ${reqNo} [${status}]`))
        .catch(err => console.log(`FAIL - ${reqNo} [${err.response && err.response.status}]: ${(err.response && err.response.data && JSON.stringify(err.response.data)) || err.message}`))
        .then(() => reqNo)
      );
    } else {
      const reqNo = await Promise.any(Object.values(inflight))
      delete inflight[reqNo]
    }
  }

}

run().catch(err => {
  console.error(err.stack);
  process.exit(1);
});
```


peterbroadhurst:
Note I've put in a bunch of fixes to the UTs, unrelated to the change. As I was getting intermittent failures on the Github actions.

Headlines of this are:
- Swapped order in a case where mocks were being set up after kicking off a go routine that might use them https://github.com/hyperledger/firefly-ethconnect/pull/175/commits/6092d5620220f2b63c5c524b502317ed0eb0845f
- Made it so that UTs wait until all `eventstream`s close before exiting, to avoid cases where we hit mocks on the background eventpoller, after the main test routine has torn them down
- In doing this, found a bug in the code where `postUpdateStream` wasn't called in error cases, even though `preUpdateStream` had been called. This left streams paused, and with a broken state (no `updateInterrupt` closed)

peterbroadhurst:
Was still seeing errors, so went a step further and replaces the `WaitGroup` logic (which had got very complicated) with just separate `chan struct{}` channels for each worker.

internal/kafka/circuitbreaker.go
```diff
@@ -0,0 +1,184 @@
...
+	gap          int64
+	bufSize      int64
+	msgCount     int64
+	msgBytes     int64
+	tripped      bool
+	lastTripTime time.Time
+	lastLogged   time.Time
+}
+
+// circuitBreaker checks that the
```
vdamle, line 63: incomplete comment

internal/kafka/circuitbreaker.go
```diff
@@ -0,0 +1,184 @@
...
+	conf  *CircuitBreakerConf
+	mux   sync.Mutex
+	state map[string]map[int32]*cbPartitionState
+}
+
+var singletonCircuitBreaker *circuitBreaker
+
+func InitCircuitBreaker(conf *CircuitBreakerConf) error {
+
+	if conf == nil || !conf.Enabled {
```
vdamle, line 74: Would be good to add a log here

internal/kafka/circuitbreaker.go
```diff
@@ -0,0 +1,184 @@
...
+	tripped      bool
+	lastTripTime time.Time
+	lastLogged   time.Time
+}
+
+// circuitBreaker checks that the
+type circuitBreaker struct {
+	conf  *CircuitBreakerConf
+	mux   sync.Mutex
+	state map[string]map[int32]*cbPartitionState
```
vdamle, line 67: Can we add a comment that describes the mapping `{topic} -> {partition} -> {partitionState}` is what it appears to be, based on reading the code further down.

internal/rest/receiptstore.go
```diff
@@ -158,8 +168,8 @@ func (r *receiptStore) writeReceipt(requestID string, receipt map[string]interfa
 		// Check if the reason is that there is a receipt already
 		existing, qErr := r.persistence.GetReceipt(requestID)
 		if qErr == nil && existing != nil {
-			log.Warnf("%s: exiting   receipt: %+v", requestID, *existing)
-			log.Warnf("%s: duplicate receipt: %+v", requestID, receipt)
+			log.Debugf("%s: exiting   receipt: %+v", requestID, *existing)
```
vdamle, line 171:
```suggestion
			log.Debugf("%s: existing   receipt: %+v", requestID, *existing)
```

internal/kafka/circuitbreaker.go
```diff
@@ -0,0 +1,184 @@
...
+	partitionState.gap = (hwm - offset)
+	if partitionState.gap <= 0 {
+		partitionState.gap = 0 // ensure positive
+	}
+	partitionState.bufSize = (partitionState.gap * partitionState.sizeEstimate)
+
+	if partitionState.tripped {
+		if partitionState.bufSize < cb.conf.resetBufferSize {
+			partitionState.tripped = false
+			cb.logState("CircuitBreakerReset", topic, partition, partitionState)
```
vdamle, line 149: Wondering if a `lastResetTime` would be useful as well in diagnosing problems with chatty/bursty submissions.

vdamle (commented):
Thanks for the fixes to the event stream update logic @peterbroadhurst , I can see it simplifies coordination logic!

Added a few suggestions to the circuit breaker portion as well for consideration.

jimthematrix (commented):
The current implementation relies on the consumer to update the gap count.

Is it ever possible for the consumer group to be stuck under certain situations while the produce continues to pump new messages into the queue?

don't know if makes sense to periodically get the HWM using the consumer interface so that the gap can continue to be tracked even if the consumer group is not able to continue processing messages...

again this is a hypothetical question as I don't know Kafka enough to know if it's possible to get into this situation

peterbroadhurst:
> don't know if makes sense to periodically get the HWM using the consumer interface so that the gap can continue to be tracked even if the consumer group is not able to continue processing messages...

This is valid feedback. The current design (per discussion in #173) is a compromise due to the complexity and overhead of making the producer also create a consumer. I will look into the feasibility, but propose doing that evaluation separate to this PR.

jimthematrix:
> ...due to the complexity and overhead of making the producer also create a consumer. ... propose doing that evaluation separate to this PR.

I see that this was already considered in the design 👍 agree that it can be treated as additional enhancement in a future PR

jimthematrix (approved):

PR #177, opened 2021-11-22
POST to /subscriptions with in-line event definition

peterbroadhurst:
FireFly will be handling the storage and management of ABI event types, so we need a base route to `POST` `/subscriptions` that allows the event definition, address etc. to be supplied in-line.


awrichar (approved):

internal/contractgateway/smartcontractgw_test.go
```diff
@@ -1688,6 +1688,41 @@ func TestDeleteSub(t *testing.T) {
...
+
+	mockSubMgr := &mockSubMgr{
+		sub: &events.SubscriptionInfo{
+			Name: "mysub",
+		},
+	}
+	var resBody events.SubscriptionInfo
+	res := testGWPathBody("POST", events.SubPathPrefix, &resBody, mockSubMgr, bytes.NewReader([]byte(`
+    {
+      "name": "mysub"
```
awrichar, line 1702: Would be nice to see the test pass in all of the possible fields (like `Event` and `Address`) just to validate the expected format for them.

peterbroadhurst: Done

vdamle (approved):

PR #193, opened 2022-01-22
Fixing contract deployment and example doc

drewmarshburn:
When trying to run the sample contract deployment payload, I was getting this error:

```
ERROR <-- POST /hook [500]: exec: "solc": executable file not found in $PATH
```

I've added the `solc` executable to the output image and then updated the `DeployContract` sample payload to be compatible with the newer version of `solc`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN make clean all
 
 FROM debian:buster-slim
 WORKDIR /ethconnect
+COPY --from=builder /usr/bin/solc /usr/bin/solc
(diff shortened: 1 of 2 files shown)


peterbroadhurst:
I am going to suggest we do not include this.
Adding a `solc` executable to the FireFly image is a valid thing for a particular project to do, but I don't think it would be the remit of the HL main project to decide to do this.

I note that neither FireFly Core or FireFly CLI are designed to make use of EthConnect for contract compilation.

Happy to revisit, but for now I'm going to close this out - fyi @nguyer @drewmarshburn

PR #195, opened 2022-01-24
Fixing Graceful Shutdown for Container Image

onelapahead:
Noticed in a K8s environment when testing https://github.com/hyperledger/firefly-helm-charts/pull/12 that ethconnect was very slow to shut down. Likely bc its not receiving the sigterm: https://hynek.me/articles/docker-signals/

> If you run your application from a shell script the regular way, your shell spawns your application in a new process and you 
> won’t receive signals from Docker.
>
> What you need to do is to tell your shell to replace itself with your application. For that exact purpose, shells have the exec > command (the similarity to the exec form before is not an accident: exec syscall).

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,12 +1,15 @@
 #!/bin/sh
 
+args=""
 if [ "$#" -eq 0 ]; then
     if [ -n "${ETHCONNECT_CONFIGFILE}" ]; then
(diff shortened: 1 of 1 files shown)


vdamle (approved):
🛑

PR #196, opened 2022-02-01
Allow passing an array of ABI methods that might emit an event, when creating a sub

peterbroadhurst:
Two features in this PR related to the `inputs: true` feature on event streams:

### 1. Include `inputSigner` in the response

Return the ethereum address of the `from` that comes back in the `eth_getTransactionByHash` when we're doing the input decoding. This is needed by the FireFly tokens connectors, and currently they have to make an inefficient extra call.

### 2. Add a list of method ABI signatures when subscribing, to avoid need for ABI to be registered

When you are creating a subscription directly, on `POST` `/subscriptions` with an inline Event definition you might still want the inputs to be resolved (the stream has `inputs: true` set).

However, there's no tie between the subscription and an an ABI in the contract registry in this case, so we need a way for you to manually specify an array of all the possible methods that might emit the event you're interested in. Otherwise, there's no way for EthConnect to match the method signature extracted from the transaction, back to a particular method ABI, in order to know how to RLP decode the input data.

This PR propose very simple extension to #177, allowing you to post the payload example below.

`POST` `/subscriptions`
```js
{
  "name": "mysub",
  "stream": "df02bbdd-7038-462e-934d-9b5c7493fcbb",
  "fromBlock": "latest",
  "address": "0x10393D362bEfa1fe52e613801D72Da358C0d76c0",
  "event": {
    "name": "stuffHappened",
    "type": "event",
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "name": "from",
        "type": "address"
      },
      {
        "indexed": true,
        "name": "val1",
        "type": "int64"
      }
    ]
  },
  "methods": [
    {
      "type": "function",
      "name": "doStuff",
      "constant": false,
      "inputs": [
        {
          "name": "val1",
          "type": "int64"
        }
      ],
      "outputs": [],
      "payable": false,
      "stateMutability": "nonpayable"
    }
  ]
}
```

### DIFF
--- a/internal/contractregistry/contractstore.go
+++ b/internal/contractregistry/contractstore.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/go-openapi/spec"
 	lru "github.com/hashicorp/golang-lru"
+	ethbinding "github.com/kaleido-io/ethbinding/pkg"
 	log "github.com/sirupsen/logrus"
 
 	ethconnecterrors "github.com/hyperledger/firefly-ethconnect/internal/errors"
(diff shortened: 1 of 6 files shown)


peterbroadhurst:
Ah, change I made to `ABILocation` affected the `contractregistry` lRU cache. Working on it

peterbroadhurst:
ok - refactored to avoid changing in `contractregistry` code

awrichar (approved):
Looks good to me; thanks for enhancing this.

Issue #198, opened 2022-02-04
A `PATCH` to the event stream is resetting all subscriptions owned by that event stream to their initial state


peterbroadhurst:
This is problematic for FireFly, which subscribes from Block 0, and uses a `PATCH` on startup to ensure the configuration of the event stream is always up to date.

```
[2022-02-04T02:57:47.779Z]  INFO --> PATCH /eventstreams/es-2259ebbd-0f87-419d-4463-29876cde5be6
[2022-02-04T02:57:47.780Z]  INFO es-2259ebbd-0f87-419d-4463-29876cde5be6: Update event stream
[2022-02-04T02:57:47.780Z]  INFO es-2259ebbd-0f87-419d-4463-29876cde5be6: Notified of an ongoing stream update, existing batch processor
[2022-02-04T02:57:47.780Z]  INFO es-2259ebbd-0f87-419d-4463-29876cde5be6: Notified of an ongoing stream update, existing event poller
[2022-02-04T02:57:47.780Z]  INFO es-2259ebbd-0f87-419d-4463-29876cde5be6: Notified of an ongoing stream update, not waiting for new events
[2022-02-04T02:57:47.800Z]  INFO sb-e1f61740-2dc7-4e1f-68ad-f0ac55e4e663:BatchPin(address,uint256,string,bytes32,bytes32,string,bytes32[]): Uninstalled filter. ok=true (%!s(<nil>))
[2022-02-04T02:57:47.800Z]  INFO <-- PATCH /eventstreams/es-2259ebbd-0f87-419d-4463-29876cde5be6 [200]
[2022-02-04T02:57:47.801Z]  INFO sb-e1f61740-2dc7-4e1f-68ad-f0ac55e4e663:BatchPin(address,uint256,string,bytes32,bytes32,string,bytes32[]): initial block height for event stream (from block): 0
[2022-02-04T02:57:47.807Z]  INFO --> GET /subscriptions
[2022-02-04T02:57:47.807Z]  INFO <-- GET /subscriptions [200]
[2022-02-04T02:57:47.819Z]  INFO sb-e1f61740-2dc7-4e1f-68ad-f0ac55e4e663:BatchPin(address,uint256,string,bytes32,bytes32,string,bytes32[]): catchup mode. Blocks 0 -> 4999
[2022-02-04T02:57:47.822Z]  INFO WS/8d27544e-a7ce-460d-6442-0621118b27dc: Connected
[...]

PR #199, opened 2022-02-04
Protect misconfiguring catchupModeBlockGap < catchupModePageSize and fix catchup

peterbroadhurst:
See #198 for the 🤕 behavior you can get if we don't protect against this.

The logic here has an implicit assumption of this rule, so that when we read a page it's never going to extend past the `blockNumber` of the header of the chain at the point we query.

Otherwise, we think:
- Gap to head 1234 is > 250 - so we're in catchup
- Read 5000 (page size) - cool we've read 5000 blocks ... but actually there were only 1234 to read!!!
- ok, now we're not in catchup any more - start reading from where we left off - but that's now skipped a whole bunch of blocks.

https://github.com/hyperledger/firefly-ethconnect/blob/f6c7fe5499d111a25d2301b84efdb8a13db988e2/internal/events/subscription.go#L223-L248

### DIFF
--- a/internal/events/eventstream.go
+++ b/internal/events/eventstream.go
@@ -404,7 +404,6 @@ func (a *eventStream) eventPoller() {
 
 	ctx := auth.NewSystemAuthContext()
 	var checkpoint map[string]*big.Int
-	blockUpdatedFilterStale := false
 	for !a.suspendOrStop() {
 		var err error
 		// Load the checkpoint (should only be first time round)
@@ -428,14 +427,12 @@ func (a *eventStream) eventPoller() {
 					blockHeight, exists := checkpoint[sub.info.ID]
 					if !exists || blockHeight.Cmp(big.NewInt(0)) <= 0 {
 						blockHeight, err = sub.setInitialBlockHeight(ctx)
-					} else {
+					} else if !sub.inCatchupMode() {
 						sub.setCheckpointBlockHeight(blockHeight)
 					}
 					if err == nil {
 						err = sub.restartFilter(ctx, blockHeight)
 					}
-					blockUpdatedFilterStale = true
-					log.Debugf("%s: Checkpoint updated due to stale sub filter: %s", sub.info.ID, blockHeight.String())
 				}
 				if err == nil {
 					err = sub.processNewEvents(ctx)
@@ -453,15 +450,17 @@ func (a *eventStream) eventPoller() {
 				i1 := checkpoint[sub.info.ID]
 				i2 := sub.blockHWM()
 
-				changed = changed || blockUpdatedFilterStale || i1 == nil || i1.Cmp(&i2) != 0
+				subChanged := i1 == nil || i1.Cmp(&i2) != 0
+				if subChanged {
+					log.Debugf("%s: New checkpoint HWM: %s", a.spec.ID, i2.String())
+				}
+				changed = changed || subChanged
 				checkpoint[sub.info.ID] = new(big.Int).Set(&i2)
 			}
 			if changed {
 				if err = a.sm.storeCheckpoint(a.spec.ID, checkpoint); err != nil {
 					log.Errorf("%s: Failed to store checkpoint: %s", a.spec.ID, err)
 				}
-				// set this to false before the next evaluation for stale subscriptions
-				blockUpdatedFilterStale = false
 			}
 		}
 		// the event poller reacts to notification about a stream update, else it starts
--- a/internal/events/logprocessor.go
+++ b/internal/events/logprocessor.go
@@ -67,11 +67,13 @@ type logProcessor struct {
 }
 
 func newLogProcessor(subID string, event *ethbinding.ABIEvent, stream *eventStream) *logProcessor {
-	return &logProcessor{
+	lp := &logProcessor{
 		subID:  subID,
 		event:  event,
 		stream: stream,
 	}
+	lp.highestDispatched.SetInt64(-1)
+	return lp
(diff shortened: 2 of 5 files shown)


peterbroadhurst:
I found we still were not writing checkpoint HWM updates while replaying in catchup mode.
There had been an attempted fix in https://github.com/hyperledger/firefly-ethconnect/pull/157, but it didn't get to the root cause - so I've backed that out.

There was a check here that architecturally was meant to address this when https://github.com/hyperledger/firefly-ethconnect/pull/130 was first implemented, but it has a special case of failing when `highestDispatched` is `0`:
https://github.com/hyperledger/firefly-ethconnect/blob/f6c7fe5499d111a25d2301b84efdb8a13db988e2/internal/events/logprocessor.go#L98

So I've just made `highestDispatched` init to `-1`.

internal/events/eventstream.go
```diff
@@ -428,14 +427,12 @@ func (a *eventStream) eventPoller() {
 					blockHeight, exists := checkpoint[sub.info.ID]
 					if !exists || blockHeight.Cmp(big.NewInt(0)) <= 0 {
 						blockHeight, err = sub.setInitialBlockHeight(ctx)
-					} else {
+					} else if !sub.inCatchupMode() {
```
peterbroadhurst, line 430: We were constantly restoring back to the checkpoint (and logging that fact) in catchup mode. No need

internal/events/eventstream.go
```diff
@@ -453,15 +450,17 @@ func (a *eventStream) eventPoller() {
 				i1 := checkpoint[sub.info.ID]
 				i2 := sub.blockHWM()
 
-				changed = changed || blockUpdatedFilterStale || i1 == nil || i1.Cmp(&i2) != 0
+				subChanged := i1 == nil || i1.Cmp(&i2) != 0
+				if subChanged {
+					log.Debugf("%s: New checkpoint HWM: %s", a.spec.ID, i2.String())
```
peterbroadhurst, line 455: Replaced other less helpful logging, to just show the one subscription that updated its checkpoint

internal/events/logprocessor.go
```diff
@@ -67,11 +67,13 @@ type logProcessor struct {
 }
 
 func newLogProcessor(subID string, event *ethbinding.ABIEvent, stream *eventStream) *logProcessor {
-	return &logProcessor{
+	lp := &logProcessor{
 		subID:  subID,
 		event:  event,
 		stream: stream,
 	}
+	lp.highestDispatched.SetInt64(-1)
```
peterbroadhurst, line 75: Key fix - start with `highestDispatched` at `-1` so compare to `0` works

internal/events/logprocessor.go
```diff
@@ -96,9 +98,8 @@ func (lp *logProcessor) getBlockHWM() big.Int {
 func (lp *logProcessor) markNoEvents(blockNumber *big.Int) {
 	lp.hwnSync.Lock()
 	if lp.highestDispatched.Cmp(&lp.blockHWM) < 0 {
-		// Nothing in-flight, its safe to update the HWM
-		lp.blockHWM.Set(blockNumber)
-		log.Debugf("%s: HWM: %s", lp.subID, lp.blockHWM.String())
+		// Nothing in-flight, its safe to update the HWM - to one after the block we're up to
+		lp.blockHWM.Set(new(big.Int).Add(blockNumber, big.NewInt(1)))
```
peterbroadhurst, line 102: Without this, each restart we were going back one block.

internal/events/submanager.go
```diff
@@ -122,6 +122,10 @@ func NewSubscriptionManager(conf *SubscriptionManagerConf, rpc eth.RPCClient, cr
 	if conf.CatchupModePageSize <= 0 {
 		conf.CatchupModePageSize = defaultCatchupModePageSize
 	}
+	if conf.CatchupModeBlockGap < conf.CatchupModePageSize {
```
peterbroadhurst, line 125: Key fix - stop users from setting invalid combination

internal/events/submanager.go
```diff
@@ -410,7 +414,7 @@ func (s *subscriptionMGR) loadCheckpoint(streamID string) (map[string]*big.Int,
 func (s *subscriptionMGR) storeCheckpoint(streamID string, checkpoint map[string]*big.Int) error {
 	cpID := checkpointIDPrefix + streamID
 	b, _ := json.MarshalIndent(&checkpoint, "", "  ")
-	log.Debugf("Storing checkpoint %s: %s", cpID, string(b))
+	log.Tracef("Storing checkpoint %s: %s", cpID, string(b))
```
peterbroadhurst, line 417: Move this to trace as we now log at Debug the individual subs above, only when they change

internal/events/eventstream.go
vdamle: Ok, this took me a bit of time to understand, so adding a note here for my future self. The checkpoint is updated in catchup mode as part of `batchComplete()` call: https://github.com/kaleido-io/firefly-ethconnect/blob/511d55a563aa7311c310ecadd4aa3040760abead/internal/events/logprocessor.go#L75-L76

internal/events/logprocessor.go
vdamle: Wow, definitely missed this the last time I worked on fixing the restart!

vdamle (approved):
Agree with reverting the previous iteration of attempted fix, @peterbroadhurst . Changes LGTM

PR #204, opened 2022-03-01
Remove newline between fields in log

vdamle:
This PR removes a newline in a log that outputs the request ID and offset. Logging on the same line makes it easier to search for an offset or a request ID in the logs to correlate the lifecycle of a request (when received, when received on the consumer side, final response sent etc.)

Signed-off-by: Vinod Damle <vinod.damle@kaleido.io>


peterbroadhurst (approved):
👍

Issue #149, opened 2021-09-16
Support for public blockchains


glesaint:
# Background

EthConnect was initially designed as a bridge into Ethereum permissioned chains. These chains usually are running a PoA/IBFT consensus algorithm, where the finality of transactions is guaranteed. These chains are usually configured with a gas price of 0, so that transaction fees are not required. Lastly, the permissioned chain provider usually offers a integrated signing service.
Users of public blockchains have similar needs for an ESB/EAI integration in many cases, which EthConnect could fulfill provided that the issues outlined above are resolved.

# Motivation

It's beneficial for the community to have a product that addresses the needs of public, and private chain users, so that developers working on the Ethereum platform have a set of tools available to them that can address the any requirement for integration.
At first sight, the work required to enable this in EthConnect does not seem to have any roadblock.

# Requirements

## Gas Price
- Public network transactions usually fetch a gas price from a 3rd party source such as ethgasstation.info for a desired confirmation time
- EthConnect should allow to send a transaction with high/medium/low speeds, or even a target confirmation time (e.g. 1mn, 15mn, 1hr)
- EthConnect should keep track of transactions confirmations, and allow to rebroadcast transactions with a higher fee when the desired confirmation speed is not achieved

## Finality
- Finality on public networks is usually defined in terms of number of block confirmations
- Transactions can enter the transaction pool, and exit it after some time (E.g. dropped by a node)
- EthConnect should not assume finality but instead allow users to define the required number of block confirmations
- For security reasons, we usually also verify the number of confirmations from multiple attesting parties

## External Signing
- EthConnect signs transactions directly currently
- On public networks, a variety of signing solutions (e.g. metamask, hardware wallet, custodian service) would be used
- If we want to continue to rely on ethconnect for ordering and nonce, EthConnect should allow to use an external signing service
- Perhaps a signing request feature should be introduced that's separate from the signing itself that could happen on another service

## Events Detection
- On public networks, transactions that have been confirmed can, in rare cases, then disappear
- These scenarios occur when a transaction is confirmed on a block that ends up being an uncle in Ethereum
- The events notification/stream should be amended so that a 'Orphaned' status is used for such transaction states, for example

# Outline
To be discussed with the community on how the above can be implemented.

peterbroadhurst:
I think there is a significant architectural point to consider, when you take a few of the above requirements into account.

[...]

PR #207, opened 2022-03-22
Support for confirmations when listening to events

peterbroadhurst:
- One of the key work items for support of public blockchains per #149 
- Adds a new `BlockConfirmationManager` (BCM) component
  - Enabled through a new `confirmations` configuration sub-section section in `openapi`
  ```yaml
  rest:
    rest-gateway:
      rpc:
        url: https://rinkeby.infura.io/v3/...
      openapi:
        catchupModeBlockGap: 50
        catchupModePageSize: 50
        confirmations:
          enabled: true               # Turn on the feature
          requiredConfirmations: 20   # The required number of confirmations to deliver the event
          includeInPayload: true      # Includes an array of the block confirmations in the event
        decimalTransactionIndex: true # Gives consistency in decimal formatting of transactionIndex
      maxTXWaitTime: 300
      maxInFlight: 100
  ```
  - Sits in-between the `LogProcessor` and the `EventStream`
  - Responsible for detecting confirmations, and only propagating events once confirmed
  - Events stay in-flight until confirmed, so checkpointing logic functions unchanged
  - Performs polling based block listening to add confirmations as soon as they arrive
  - Handles confirmations of historic events when replaying
  - Handles forks that invalidate at any point in the confirmation list
  - Block caching avoids excessive JSON/RPC calls against the node
  - Tested with the Infura gateway for the Rinkeby network
- Updates handling of `removed` events
  - Processed by BCM when enabled
  - Ignored otherwise
- Adds migration for `blockedReryDelaySec` config typo to `blockedRetryDelaySec`
- Adds a `hexFormatTransactionIndex` option that can be set to `false` to disable the current behavior where `transactionIndex` is delivered as `0x01` etc. whereas `blockNumber`/`logIndex` are delivered as the string `"1"`
   - The default has not been changed to avoid impact on production apps


.vscode/settings.json
```diff
@@ -1,3 +1,13 @@
 {
-  "go.lintTool": "golangci-lint"
+  "go.lintTool": "golangci-lint",
+  "cSpell.words": [
```
vdamle, line 3: 👍🏼

internal/events/block_confirmations.go
```diff
@@ -0,0 +1,520 @@
+// Copyright 2019 Kaleido
```
vdamle, line 1:
```suggestion
// Copyright 2022 Kaleido
```

internal/events/block_confirmations.go
```diff
@@ -0,0 +1,520 @@
...
+		requiredConfirmations: defaultConfirmations,
+		blockCacheSize:        defaultBlockCacheSize,
+		pollingInterval:       defaultPollingInterval,
+		eventQueueLength:      defaultEventQueueLength,
+	}
+	if conf.RequiredConfirmations != nil {
+		intConf.requiredConfirmations = *conf.RequiredConfirmations
+	}
+	if conf.BlockCacheSize != nil {
+		intConf.blockCacheSize = *conf.BlockCacheSize
```
vdamle, line 141: should we enforce a safe upper bound (no idea what that should be though!)?

internal/events/block_confirmations.go
```diff
@@ -0,0 +1,520 @@
...
+			continue
+		}
+
+		// Process each new block
+		bcm.processBlockHashes(blockHashes)
+
+		// Process any new notifications - we do this at the end, so it can benefit
+		// from knowing the latest highestBlockSeen
+		if err := bcm.processNotifications(notifications); err != nil {
+			bcm.log.Errorf("Failed processing notifications: %s", err)
```
vdamle, line 321:
```suggestion
			bcm.log.Errorf("Failed to process notifications: %s", err)
```
along the lines of other messages..

internal/events/block_confirmations.go
```diff
@@ -0,0 +1,520 @@
...
+	// Add the event
+	eventKey := bcm.keyForEvent(event)
+	pending := &pendingEvent{
+		key:           bcm.keyForEvent(event),
+		added:         time.Now(),
+		confirmations: make([]*blockInfo, 0, bcm.requiredConfirmations),
+		event:         event,
+		eventStream:   eventStream,
+	}
+	bcm.pending[eventKey] = pending
```
vdamle, line 378: Not requesting as part of this PR given I'm providing this comment at the last minute, it would be good to have a follow on that wraps updates to `bcm.pending` with a RWMutex - I'm assuming there's a high likelihood of concurrent updates.

internal/events/block_confirmations.go
```diff
@@ -0,0 +1,520 @@
...
+		bcm.dispatchConfirmed(c)
+	}
+
+}
+
+// dispatchConfirmed drive the event stream for any events that are confirmed, and prunes the state
+func (bcm *blockConfirmationManager) dispatchConfirmed(confirmed *pendingEvent) {
+	eventKey := bcm.keyForEvent(confirmed.event)
+	bcm.log.Infof("Confirmed with %d confirmations event=%s", len(confirmed.confirmations), eventKey)
+	delete(bcm.pending, eventKey)
```
vdamle, line 453: Wouldn't we have already removed the key from `bcm.pending` since we moved this event from `bcm.pending` to `confirmed`?

internal/events/eventstream.go
```diff
@@ -231,9 +254,25 @@ func (a *eventStream) preUpdateStream() error {
 	close(a.updateInterrupt)
 	a.batchCond.Broadcast()
 	a.batchCond.L.Unlock()
+
+	a.drainBlockConfirmationManager()
+
 	return nil
 }
 
+func (a *eventStream) drainBlockConfirmationManager() {
```
vdamle, line 263: should we purge the cache here?

internal/events/block_confirmations_test.go
```diff
@@ -0,0 +1,802 @@
+// Copyright 2019 Kaleido
```
vdamle, line 1:
```suggestion
// Copyright 2022 Kaleido
```

internal/events/block_confirmations.go
peterbroadhurst: I guess we haven't elsewhere on these. I propose no is ok for now

internal/events/block_confirmations.go
peterbroadhurst: Agree a good candidate to look at in the future, and deferrable.

internal/events/block_confirmations.go
peterbroadhurst: Agreed reviewing code

internal/events/eventstream.go
peterbroadhurst: It's intentional we don't here, because the drain is only scoped to one event stream.
Whereas the cache is across all event streams, based on blocks.

vdamle (approved):

Issue #205, opened 2022-03-10
`concurrent map writes` panic in ethconnect


peterbroadhurst:
```
firefly_e2e_ethconnect_1  | [2022-03-10T17:54:21.107Z]  INFO Created subscription ID:sb-aeb1bb92-e19e-4a6a-7d21-3090683649e5 name:token:N2:ApprovalForAll topic:0x17307eab39ab6107e8899845ad3d59bd9653f200f220920489ca2b5937696c31
firefly_e2e_ethconnect_1  | fatal error: [2022-03-10T17:54:21.106Z]  INFO <-- POST /contracts/erc1155/TransferBatch [200]
firefly_e2e_ethconnect_1  | concurrent map writes
firefly_e2e_ethconnect_1  | [2022-03-10T17:54:21.107Z] DEBUG <-- {"created":"2022-03-10T17:54:21Z","id":"sb-b39cbe9d-9d90-48cd-7e81-d0b763d138c4","path":"/subscriptions/sb-b39cbe9d-9d90-48cd-7e81-d0b763d138c4","name":"token:N2:TransferBatch","stream":"es-f63c7e42-9545-4116-7274-460b66672a36","filter":{"address":["0xd45aeae96f9446b9c8bcd0e655ab902d22fa3b5c"],"topics":[["0x4a39dc06d4c0dbc64b70af90fd698a233a518aa5d07e595d983b8c0526c8f7fb"]]},"event":{"type":"event","name":"TransferBatch","inputs":[{"name":"operator","type":"address","internalType":"address","indexed":true},{"name":"from","type":"address","internalType":"address","indexed":true},{"name":"to","type":"address","internalType":"address","indexed":true},{"name":"ids","type":"uint256[]","internalType":"uint256[]"},{"name":"values","type":"uint256[]","internalType":"uint256[]"}],"outputs":null},"fromBlock":"38","abi":{"type":2,"name":"712ea57a-3ac5-47f2-4c1f-331e31752b72"}}
firefly_e2e_ethconnect_1  | [2022-03-10T17:54:21.107Z]  INFO <-- POST /contracts/erc1155/TransferSingle [200]
[...]

PR #208, opened 2022-03-30
Fix concurrent map write when adding multiple subscriptions in parallel

nguyer:
Resolves https://github.com/hyperledger/firefly-ethconnect/issues/205

### DIFF
--- a/internal/events/submanager.go
+++ b/internal/events/submanager.go
@@ -493,9 +493,9 @@ func (s *subscriptionMGR) recoverSubscriptions() {
 			if err != nil {
 				log.Errorf("Failed to recover subscription '%s': %s", subInfo.ID, err)
 			} else {
-				s.subscriptionsMutex.RLock()
+				s.subscriptionsMutex.Lock()
 				s.subscriptions[subInfo.ID] = sub
-				s.subscriptionsMutex.RUnlock()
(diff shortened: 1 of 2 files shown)


internal/events/submanager.go
```diff
@@ -86,14 +87,15 @@ type SubscriptionManagerConf struct {
...
-	wsChannels    ws.WebSocketChannels
+	conf               *SubscriptionManagerConf
+	db                 kvstore.KVStore
+	rpc                eth.RPCClient
+	subscriptions      map[string]*subscription
+	streams            map[string]*eventStream
+	closed             bool
+	cr                 contractregistry.ContractResolver
+	wsChannels         ws.WebSocketChannels
+	subscriptionsMutex sync.Mutex
```
vdamle, line 98: For concurrent access that is safe for both Reads & Writes, I suggest we use [RWMutex](https://pkg.go.dev/sync#RWMutex) and protect Read operations as well from concurrent access. Ex: https://github.com/hyperledger/firefly-ethconnect/blob/main/internal/events/submanager.go#L134

vdamle (changes requested):
As noted in my comment, we should ensure Read operations are also safe.

vdamle (approved):

peterbroadhurst (approved):

PR #216, opened 2022-05-05
Replace dep on fftm+core with new firefly-common package

peterbroadhurst:
We ended up pulling in a huge tree of dependencies  by the definitions for the FFCAPI being FireFly Transaction Manager, which depends then on FireFly Core for types. Now everything has moved to FireFly Common, we have a much slimmer set of deps to pull in.

Also renamed `ffc` package to `ffcapiconnector` on feedback that `ffc` was too vague.

Depends on (go.mod pulls these in directly):
- https://github.com/hyperledger/firefly-common/pull/4
- https://github.com/hyperledger/firefly/pull/791

Also see:
- https://github.com/hyperledger/firefly-transaction-manager/pull/6


awrichar (approved):

JMP2590 (commented):
👍

PR #217, opened 2022-05-11
Add fly-id based idempotency for LevelDB (as implemented for MongoDB)

peterbroadhurst:
The `fly-id` (/`kld-id`) query param for idempotent submission of transactions, using an external ID combined with `fly-acktype=receipt` (/`kld-acktype`), functions based on the DB requiring uniquness.

LevelDB doesn't have this as a first class construct, so we need to implement it at the ethconnect code layer.
 
Solving this e2e meant implementing an `overwrite` option through the stack, to pick between "keep trying to insert, with overwrite" functionality when storing the final reply, vs. "try inserting, and fail immediately" functionality when storing the initial `acktype=receipt`.

I also found an unrelated issue with the in-memory receipt store, where it was not retaining _any_ history by default when configured via YAML. A real pain for FireFly CLI based environments.

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -511,8 +511,8 @@ var (
 	// FFCBlockNotAvailable is returned when a receipt is not found
 	FFCBlockNotAvailable = e(100218, "Block not available")
 
-	// ReceiptStoreLevelDBKeyNotUnique idempotency check failed on request ID
-	ReceiptStoreLevelDBKeyNotUnique = e(100219, "Request ID is not unique")
+	// ReceiptStoreKeyNotUnique non-unique request ID
+	ReceiptStoreKeyNotUnique = e(100219, "Request ID is not unique")
 )
 
 type EthconnectError interface {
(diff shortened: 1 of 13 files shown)


jimthematrix (approved):

internal/errors/errors.go
```diff
@@ -510,6 +510,9 @@ var (
 	FFCRequestTypeNotImplemented = e(100217, "FFCAPI request '%s' not currently supported")
 	// FFCBlockNotAvailable is returned when a receipt is not found
 	FFCBlockNotAvailable = e(100218, "Block not available")
+
+	// ReceiptStoreLevelDBKeyNotUnique idepotency check failed on request ID
```
vdamle, line 514:
```suggestion
	// ReceiptStoreLevelDBKeyNotUnique idempotency check failed on request ID
```
..in the spirit of the other change in this PR ;)

vdamle (approved):

internal/errors/errors.go
peterbroadhurst: Guess I can only correct that spelling once :)

PR #218, opened 2022-05-16
Updates for parallel processing

peterbroadhurst:
Three related changes from analysis of some performance tests:

1. The `sendConcurrency` setting is not currently providing the full concurrency support it was intended to. This is because the `concurrencySlots` channel was being allocated in the constructor of the `TxnProcessor` which is before `SetConf` is called. So we need to allocated it in `Init()`. This results in us using pipelining between the `Send` and in-flight nonce allocation, but not performing multiple `Send` calls in parallel. 

2. In the case that a `Send` fails due to an intermittent error, such as a timeout, this is problem for end-users because they do not know if the transaction was submitted or not. However, in cases where EthConnect is in control of the `nonce` it is safe for EthConnect to retry. So this PR introduces an option for retries, and enables these by default (with a small number - 5) to deal with glitches without pushing those errors all the way back to the requester.

3. Currently the addressbook lookup is performed on the single-threaded processing, before we dispatch a concurrent worker to the `inflight` action. As there is a JSON/RPC healthcheck call that is part of this (`net_version`) this is inefficient. So this PR moves it to the concurrent worker, after `inflight` is generated.

### DIFF
--- a/internal/tx/txnprocessor.go
+++ b/internal/tx/txnprocessor.go
@@ -34,7 +34,11 @@ import (
 )
 
 const (
-	defaultSendConcurrency = 1
+	defaultSendConcurrency   = 1
+	defaultSendRetryMax      = 5
+	defaultSendRetryMinDelay = 500 * time.Millisecond
+	defaultSendRetryMaxDelay = 5 * time.Second
+	defaultSendRetryFactor   = 2.0
 )
 
 // TxnProcessor interface is called for each message, as is responsible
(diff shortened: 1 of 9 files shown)


internal/tx/txnprocessor.go
```diff
@@ -133,6 +150,25 @@ func (p *txnProcessor) Init(rpc eth.RPCClient) {
 	if p.conf.HDWalletConf.URLTemplate != "" {
 		p.hdwallet = newHDWallet(&p.conf.HDWalletConf)
 	}
+	p.concurrencySlots = make(chan bool, p.conf.SendConcurrency)
```
vdamle, line 153: After a discussion with @peterbroadhurst , I note that this needs to be setup in `Init()` because of the way this function and the caller of this function `NewKafkaBridge()` is called: https://github.com/hyperledger/firefly-ethconnect/blob/442cfb353a6e6f265a358bef2e39c07e0e1f2c0d/cmd/ethconnect.go#L172-L173

The config is set _after_ the constructor is invoked, so setting it in the constructor doesn't have the intended effect.

PR #222, opened 2022-06-07
return a slice of length 0 - signed-off

ShubhamSolanki24:
Signed-off-by: ShubhamSolanki24 <shubham24071999@gmail.com>


awrichar:
Looks like there is a unit test failure now.

PR #224, opened 2022-07-13
Do not require the filter id to be numeric

peterbroadhurst:
The JSON/RPC spec has previously stated that the result of an `eth_newFitler` will be a `QUANTITY`, and that is so in Besu/Go-ethereum (and derivatives). However, we've found that there are EVM chain node implementations return other values, such as a UUID encoded to a string.

EthConnect doesn't have any reason to rely on it being a numeric, as it's just an opaque value that stores and sends back on future calls.

The latest generated official docs (at the current point less detailed than the previous wiki based docs) Ethereum JSON/RPC spec docs state `string` as the result:
https://ethereum.github.io/execution-apis/api-documentation/

So this PR moves EthConnect to just use a string

### DIFF
--- a/internal/events/block_confirmations.go
+++ b/internal/events/block_confirmations.go
@@ -37,7 +37,7 @@ type blockConfirmationManager struct {
 	ctx                   context.Context
 	cancelFunc            func()
 	log                   *log.Entry
-	filterID              ethbinding.HexBigInt
+	filterID              string
 	filterStale           bool
 	rpc                   eth.RPCClient
 	requiredConfirmations int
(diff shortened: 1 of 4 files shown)


jimthematrix (approved):

vdamle (approved):

peterbroadhurst:
Dropping back to draft while I work through some other related changes I'd like to propose alongside this.

peterbroadhurst:
ok - it's actually just the filter change needed. The other issue I was chasing down was in the associated JSON/RPC server runtime.

PR #227, opened 2022-08-25
Add additional idempotence check to cover Kafka server restart, while EthConnect stays running

peterbroadhurst:
EthConnect works on an "at target" nonce allocation (see comparision with "at source" in
[readme of FFTM](https://github.com/hyperledger/firefly-transaction-manager/blob/main/README.md#nonce-management)), meaning it gives an "at least once" delivery assurance to the blockchain, backed by Apache Kafka.

In the case that the REST API Gateway is tuned for high performance, with an in-flight count in the many hundreds, and the Kafka servers are restarted, the consumer groups might redeliver many messages. This is undesirable.

EthConnect already has the concept of an idempotency key, on the front-side of the REST API Gateway, using `ackmode` as added in #175 to get an immediate receipt, combined with supplying your own custom ID. However, that is only checked in the REST API Gateway boundary layer today.

See the following diagram as a reference, showing how a Kafka at-least-once redelivery results in this duplication:
EthConnect at-least-once issue (1)

This PR proposes adding an additional idempotency check, at the point we receive the message from Kafka. Note this does not change the fundamental nature of the "at target" architecture from being at-least-once, and in some failure scenarios (for full idempotent delivery e2e with Ethereum nonces you would need the "at source" ordering architecture of https://github.com/hyperledger/firefly-evmconnect based on the new FFTM architecture).

But this PR does protect against something like a planned HA rolling restart of a Kafka cluster, from causing redelivery.

ethconnect_kafka_idempotency_enhancement

The new check is only enabled when:
- `fly-ackmode=receipt` has been specified in the transaction submission via the APIs described in #175 
- The REST API Gateway and the Kafka<->Ethereum module are co-located in the same address space

The check covers two scenarios:
1. The transaction is already in-flight in TX Processor when the redelivery occurs
2. The transaction has already been assigned a transaction hash when the redelivery occurs

> One complexity in the change, was making it so the two different components could both access the receipt store. For that I moved out a new package called `receipts`.


internal/tx/txnprocessor.go
```diff
@@ -235,8 +249,55 @@ func (p *txnProcessor) resolveSigner(from string) (signer eth.TXSigner, err erro
...
+		if alreadyInflight.msgID == inflight.msgID {
+			log.Warnf("Kafka redelivery of message already inflight: %s", inflight.msgID)
+			// We don't send a new reply here - special nil return
+			return false, nil
+		}
+	}
+	// Then check LevelDB - we should find the entry
+	r, err := p.receiptStore.GetReceipt(inflight.msgID)
+	if err != nil {
+		return false, err
```
vdamle, line 269: Do we need to add a richer return status to account for transient receipt store access issues? I'm wondering what the implication of indicating a false negative is for the application. Here, we are saying "we don't know if there is a receipt with this ID or not, assume this is _not_ idempotent".

peterbroadhurst: I did track through the result of this error return, and it will actually end up coming back as an error reply in Kafka, which would overwrite any "good" reply if there was one that was earlier.

I couldn't think of a better option here:
-  Infinite retry felt wrong under the lock
-  Silently ignoring felt wrong, because we can't sure sure an event went back at all in that case

peterbroadhurst: Thanks for this @vdamle - I've added a more descriptive error, but am open to other suggestions too

vdamle: Perfect, looks great now.

vdamle (approved):

PR #229, opened 2022-09-01
Remove early FFCAPI implementation not it is replaced by EVMConnect

peterbroadhurst:
See https://github.com/hyperledger/firefly-evmconnect and https://github.com/hyperledger/firefly-transaction-manager


jimthematrix (approved):
🔪 🔢

PR #231, opened 2022-10-12
Cleanup migrated contract JSON with full path

peterbroadhurst:
The cleanup code for migrated JSON files failed to delete the file (so would run again on each restart).


vdamle (approved):
🗑️

Issue #133, opened 2021-07-30
Arg packing for an empty slice actually returns an array


awrichar:
When packing the args for a `bytes` value that should be converted to a slice, one path actually returns a 0-length array instead of a slice, which results in the error reported for the "data" parameter here:
```
[2021-07-30T17:40:03.403Z] DEBUG Parsing args for function: function foo3(address[] to, bytes data) returns()
[2021-07-30T17:40:03.403Z] DEBUG Arg 0 requiredType: address[]
[2021-07-30T17:40:03.403Z] DEBUG Arg 0 value: [0xE57D3Fc7774DC54F1C1D618196FB4aFB933CE613] (type=[]common.Address)
[2021-07-30T17:40:03.403Z] DEBUG Arg 1 requiredType: bytes
[2021-07-30T17:40:03.403Z] DEBUG Arg 1 value: [] (type=[0]uint8)
[2021-07-30T17:40:03.403Z] ERROR Attempted to pack args [[0xE57D3Fc7774DC54F1C1D618196FB4aFB933CE613] []]: Packing arguments for method 'foo3': abi: cannot use array as type slice as argument
```

See here:
https://github.com/hyperledger-labs/firefly-ethconnect/blob/d7fd3fd213c85d164e9a610708ec912a55175969/internal/eth/txn.go#L623

awrichar:
See https://github.com/hyperledger/firefly-ethconnect/pull/222

PR #233, opened 2023-05-04
Differentiate bytes vs. fixed bytes in encoding parameters

jimthematrix:
fixes https://github.com/hyperledger/firefly-ethconnect/issues/133

for a `bytes` parameter, to input an empty value, `0x` is the expected value. It fails the packing step and returns the error as described in the issue above.

This fixes that issue by differentiating bytes vs. fixed length bytes, and returns a slice vs. fixed length array respectively


peterbroadhurst (approved):
👍

PR #239, opened 2023-11-09
[subscriptions] Fix for Validating URLs

hfuss:
We've observed a webhook subscription can be configured with URL which resolves an empty address (without throwing earlier errors):
```
panic: runtime error: index out of range [0] with length 0

goroutine 66 [running]:
github.com/hyperledger/firefly-ethconnect/internal/events.(*eventStream).isAddressUnsafe(0x11c4458, 0x1217e18)
	/kaleido-io/ethconnect/internal/events/eventstream.go:767 +0xd4
```

This is meant to treat this edge case then as an "unsafe address" so Ethconnect does not 1) panic and 2) the URL / hostname will be bubbled up as an error.

### DIFF
--- a/internal/events/eventstream.go
+++ b/internal/events/eventstream.go
@@ -764,7 +764,8 @@ func (a *eventStream) performActionWithRetry(batchNumber uint64, events []*event
 func (a *eventStream) isAddressUnsafe(ip *net.IPAddr) bool {
 	ip4 := ip.IP.To4()
 	return !a.allowPrivateIPs &&
-		(ip4[0] == 0 ||
+		(len(ip4) < 1 ||
+			ip4[0] == 0 ||
 			ip4[0] >= 224 ||
 			ip4[0] == 127 ||
 			ip4[0] == 10 ||
--- a/internal/events/webhooks.go
+++ b/internal/events/webhooks.go
@@ -54,14 +54,8 @@ func newWebhookAction(es *eventStream, spec *webhookActionInfo) (*webhookAction,
 func (w *webhookAction) attemptBatch(batchNumber, attempt uint64, events []*eventData) error {
 	// We perform DNS resolution before each attempt, to exclude private IP address ranges from the target
 	esID := w.es.spec.ID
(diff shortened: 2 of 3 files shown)


hfuss:
We determined this occurs when if the `url` of a subscription does not contain the protocol i.e. `http://` or `https://`. It feels like we should possibly have validation for that to prevent this from occurring.

Or bc we are parsing an invalid URL and ignoring the error: https://github.com/hyperledger/firefly-ethconnect/blob/main/internal/events/webhooks.go#L57

hfuss:
> Or bc we are parsing an invalid URL and ignoring the error: https://github.com/hyperledger/firefly-ethconnect/blob/main/internal/events/webhooks.go#L57

I've opted for handling this error as I believe this shouldn't be so optimistic since the subscription API does not validate the URL on creation / updates. Will defer adding such API validation for another time. This PR should be ready for review @peterbroadhurst @nguyer

internal/events/eventstream.go
```diff
@@ -764,7 +764,8 @@ func (a *eventStream) performActionWithRetry(batchNumber uint64, events []*event
 func (a *eventStream) isAddressUnsafe(ip *net.IPAddr) bool {
 	ip4 := ip.IP.To4()
 	return !a.allowPrivateIPs &&
-		(ip4[0] == 0 ||
+		(len(ip4) < 1 ||
```
hfuss, line 767: Safeguard to prevent panics and instead treat an empty addr as unsafe

internal/events/webhooks.go
```diff
@@ -114,3 +108,20 @@ func (w *webhookAction) attemptBatch(batchNumber, attempt uint64, events []*even
 	}
 	return err
 }
+
+func (w *webhookAction) validateURL() (*url.URL, *net.IPAddr, error) {
+	u, err := url.Parse(w.spec.URL)
```
hfuss, line 113: We know handle this error to late validate if what was provided on the subscription is in fact a URL and not something like a hostname or arbitrary string

peterbroadhurst (approved):

PR #243, opened 2024-03-01
Fix docker build and add to PR checks

nguyer:
These changes fix a couple issue with the Dockerfile. They also add the docker build to the PR checks so we don't end up merging something that means we can't build images anymore.


build_log.txt
awrichar, line 1: Did you mean to ignore this?

awrichar (approved):

build_log.txt
nguyer: Yep. Force pushed to make it disappear. Thanks.

